PR #1, opened 2015-08-25
Add ejson_body

tonysun83:
When loading design docs with the ID, we want to add ejson_body
to the list of options so the body is converted to term from binary.
This makes dreyfus compatible with COUCHDB2.0


rnewson:
+1

PR #3, opened 2015-09-02
Handle errors when opening an index

### LINKED ISSUE #2: Handle open errors in dreyfus_index_manager
PR incoming. PR's with numbered branch names doesn't work so hot in GitHub then eh?

davisp:
Previous if the anonymous opening process died it would cause
dreyfus_index_manager to exit and kill all other dreyfus_index
processes. This lead to bad things happening in multiple places.

This fix just tracks the anonymous opener processs in the ?BY_PID table
so we know about it when it dies. The exit status is then reused the
same as would have been for an error returned when opening the index.

Fixes: #2


rnewson:
+1

PR #6, opened 2016-05-10
Add new metric to track the "search request" latency.

brkolla:
This metric is different from what we track in clouseau. This basically will track
the overall time it took for the search request, where as the one in Clouseau will
only track the search latency at the shard level.

This will come in handy for tracking the search latency issues as seen  by the end users
and also allow us to add pager triggers based on this.

### DIFF
--- a/priv/stats_descriptions.cfg
+++ b/priv/stats_descriptions.cfg
@@ -11,6 +11,10 @@
 %% the License.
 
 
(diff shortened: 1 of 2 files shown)


rnewson:
+1

rnewson:
note, this was broken as it changed the result of the function, fixed in https://github.com/cloudant-labs/dreyfus/pull/7.

PR #7, opened 2016-06-08
Fix http handler response

eiri:
CouchDB's `chttpd:handle_request/1` expects to receive the response object from the end-point handlers so it can pass it to `after_request` epi hook. Function `handle_search_req` was returning `ok` after calling `couch_stats:update_histogram/2`, leading to chttpd's handler crash.


rnewson:
+1

PR #8, opened 2016-07-12
Don't crash on invalid ddoc

eiri:
This addresses an issue with invalid design doc crashing index updater with `case_clause` when queried index set to either not an object or object with a missing parameter "index"

BugzID: 69570


rnewson:
+1

PR #10, opened 2016-11-22
Tolerate open_int failure

rnewson:
If a database is deleted very soon after creation, it's possible for
the open_int call to return {not_found, no_db_file}. Expect this and
error cleanly out of gen_server initialisation.

BugzID: 77650


eiri:
+1

PR #11, opened 2016-11-30
Support index purge in dreyfus.

chguocloudant:
- use couch_db:fold_purged_docs to get list of documents to be purged
- update and delete index in clouseau for purged document
- write test logic to cover changed code

### DIFF
--- a/src/dreyfus_index_updater.erl
+++ b/src/dreyfus_index_updater.erl
@@ -69,29 +69,10 @@ update(IndexPid, Index) ->
 load_docs(FDI, {I, IndexPid, Db, Proc, Total, LastCommitTime, ExcludeIdRevs}=Acc) ->
     couch_task_status:update([{changes_done, I}, {progress, (I * 100) div Total}]),
     DI = couch_doc:to_doc_info(FDI),
-    #doc_info{id=Id, high_seq=Seq, revs=[#rev_info{rev=Rev, deleted=Del}|_]} = DI,
+    #doc_info{id=Id, high_seq=Seq, revs=[#rev_info{rev=Rev}|_]} = DI,
     case lists:member({Id, Rev}, ExcludeIdRevs) of
         true -> ok;
-        false ->
-            case Del of
-                true ->
-                    ok = clouseau_rpc:delete(IndexPid, Id);
-                false ->
-                    case lists:member({Id, Rev}, ExcludeIdRevs) of
-                        true -> ok;
-                        false ->
-                            {ok, Doc} = couch_db:open_doc(Db, DI, []),
-                            Json = couch_doc:to_json_obj(Doc, []),
-                            [Fields|_] = proc_prompt(Proc, [<<"index_doc">>, Json]),
-                            Fields1 = [list_to_tuple(Field) || Field <- Fields],
-                            case Fields1 of
-                                [] ->
-                                    ok = clouseau_rpc:delete(IndexPid, Id);
-                                _  ->
-                                    ok = clouseau_rpc:update(IndexPid, Id, Fields1)
-                            end
-                    end
-            end
+        false -> update_or_delete_index(IndexPid, Db, DI, Proc)
     end,
     %% Force a commit every minute
     case timer:now_diff(Now = now(), LastCommitTime) >= 60000000 of
@@ -105,39 +86,45 @@ load_docs(FDI, {I, IndexPid, Db, Proc, Total, LastCommitTime, ExcludeIdRevs}=Acc
 purge_index(Db, IndexPid, Index) ->
     {ok, IdxPurgeSeq} = clouseau_rpc:get_purge_seq(IndexPid),
     Proc = get_os_process(Index#index.def_lang),
-    true = proc_prompt(Proc, [<<"add_fun">>, Index#index.def]),
-    FoldFun = fun(PurgeSeq, {Id, _Revs}, Acc) ->
-        case couch_db:get_full_doc_info(Db, Id) of
-            not_found ->
-                ok = clouseau_rpc:delete(IndexPid, Id),
-                Acc0 = Acc;
-            FDI ->
-                DI = couch_doc:to_doc_info(FDI),
-                #doc_info{id=Id, revs=[#rev_info{rev=Rev, deleted=Del}|_]} = DI,
-                case lists:member({Id, Rev}, Acc) of
-                    true ->
-                        Acc0 = Acc;
-                    false ->
-                        case Del of
-                            true ->
-                                ok = clouseau_rpc:delete(IndexPid, Id);
-                            false ->
-                                {ok, Doc} = couch_db:open_doc(Db, DI, []),
-                                Json = couch_doc:to_json_obj(Doc, []),
-                                [Fields|_] = proc_prompt(Proc, [<<"index_doc">>, Json]),
-                                Fields1 = [list_to_tuple(Field) || Field <- Fields],
-                                case Fields1 of
-                                    [] ->
-                                        ok = clouseau_rpc:delete(IndexPid, Id);
-                                    _  ->
-                                        ok = clouseau_rpc:update(IndexPid, Id, Fields1)
-                                end
-                        end,
-                        lists:keydelete(Id, 1, Acc),
-                        Acc0 = lists:append([{Id, Rev}], Acc)
-                end
+    try
+        true = proc_prompt(Proc, [<<"add_fun">>, Index#index.def]),
+        FoldFun = fun(PurgeSeq, {Id, _Revs}, Acc) ->
+            Acc0 = case couch_db:get_full_doc_info(Db, Id) of
+                not_found ->
+                    ok = clouseau_rpc:delete(IndexPid, Id),
+                    Acc;
+                FDI ->
+                    DI = couch_doc:to_doc_info(FDI),
+                    #doc_info{id=Id, revs=[#rev_info{rev=Rev}|_]} = DI,
+                    case lists:member({Id, Rev}, Acc) of
+                        true -> Acc;
+                        false ->
+                            update_or_delete_index(IndexPid, Db, DI, Proc),
+                            [{Id, Rev} | Acc]
+                    end
+            end,
+            clouseau_rpc:set_purge_seq(IndexPid, PurgeSeq),
+            {ok, Acc0}
         end,
-        clouseau_rpc:set_purge_seq(IndexPid, PurgeSeq),
-        {ok, Acc0}
-    end,
-    couch_db:fold_purged_docs(Db, IdxPurgeSeq, FoldFun, [], []).
+        couch_db:fold_purged_docs(Db, IdxPurgeSeq, FoldFun, [], [])
+    after
+        ret_os_process(Proc)
+    end.
+
+update_or_delete_index(IndexPid, Db, DI, Proc) ->
+    #doc_info{id=Id, revs=[#rev_info{deleted=Del}|_]} = DI,
+    case Del of
+        true ->
+            ok = clouseau_rpc:delete(IndexPid, Id);
+        false ->
+            {ok, Doc} = couch_db:open_doc(Db, DI, []),
+            Json = couch_doc:to_json_obj(Doc, []),
+            [Fields|_] = proc_prompt(Proc, [<<"index_doc">>, Json]),
+            Fields1 = [list_to_tuple(Field) || Field <- Fields],
+            case Fields1 of
+                [] ->
+                    ok = clouseau_rpc:delete(IndexPid, Id);
+                _  ->
+                    ok = clouseau_rpc:update(IndexPid, Id, Fields1)
+            end
+    end.
--- a/src/dreyfus_index_updater.erl
+++ b/src/dreyfus_index_updater.erl
@@ -31,6 +31,14 @@ update(IndexPid, Index) ->
     erlang:put(io_priority, {view_update, DbName, IndexName}),
     {ok, Db} = couch_db:open_int(DbName, []),
     try
+        {ok, IdxPurgeSeq} = clouseau_rpc:get_purge_seq(IndexPid),
+        FoldFun = fun(PurgeSeq, {Id, _Revs}, Acc) ->
+            clouseau_rpc:delete(IndexPid, Id),
+            clouseau_rpc:set_purge_seq(IndexPid, PurgeSeq),
+            {ok, Acc}
+        end,
+        couch_db:fold_purged_docs(Db, IdxPurgeSeq, FoldFun, nil, []),
+
         %% compute on all docs modified since we last computed.
         TotalChanges = couch_db:count_changes_since(Db, CurSeq),
 
--- a/src/dreyfus_index_updater.erl
+++ b/src/dreyfus_index_updater.erl
@@ -31,7 +31,7 @@ update(IndexPid, Index) ->
     erlang:put(io_priority, {view_update, DbName, IndexName}),
     {ok, Db} = couch_db:open_int(DbName, []),
     try
-        ok = purge_index(Db, IndexPid, Index),
+        {ok, ExcludeIdRevs} = purge_index(Db, IndexPid, Index),
         %% compute on all docs modified since we last computed.
         TotalChanges = couch_db:count_changes_since(Db, CurSeq),
 
@@ -53,7 +53,7 @@ update(IndexPid, Index) ->
         try
             true = proc_prompt(Proc, [<<"add_fun">>, Index#index.def]),
             EnumFun = fun ?MODULE:load_docs/2,
-            Acc0 = {0, IndexPid, Db, Proc, TotalChanges, now()},
+            Acc0 = {0, IndexPid, Db, Proc, TotalChanges, now(), ExcludeIdRevs},
 
             {ok, _} = couch_db:fold_changes(Db, CurSeq, EnumFun, Acc0, []),
 
@@ -66,21 +66,31 @@ update(IndexPid, Index) ->
         couch_db:close(Db)
     end.
 
-load_docs(FDI, {I, IndexPid, Db, Proc, Total, LastCommitTime}=Acc) ->
+load_docs(FDI, {I, IndexPid, Db, Proc, Total, LastCommitTime, ExcludeIdRevs}=Acc) ->
     couch_task_status:update([{changes_done, I}, {progress, (I * 100) div Total}]),
     DI = couch_doc:to_doc_info(FDI),
-    #doc_info{id=Id, high_seq=Seq, revs=[#rev_info{deleted=Del}|_]} = DI,
-    case Del of
-        true ->
-            ok = clouseau_rpc:delete(IndexPid, Id);
+    #doc_info{id=Id, high_seq=Seq, revs=[#rev_info{rev=Rev, deleted=Del}|_]} = DI,
+    case lists:member({Id, Rev}, ExcludeIdRevs) of
+        true -> ok;
         false ->
-            {ok, Doc} = couch_db:open_doc(Db, DI, []),
-            Json = couch_doc:to_json_obj(Doc, []),
-            [Fields|_] = proc_prompt(Proc, [<<"index_doc">>, Json]),
-            Fields1 = [list_to_tuple(Field) || Field <- Fields],
-            case Fields1 of
-                [] -> ok = clouseau_rpc:delete(IndexPid, Id);
-                _  -> ok = clouseau_rpc:update(IndexPid, Id, Fields1)
+            case Del of
+                true ->
+                    ok = clouseau_rpc:delete(IndexPid, Id);
+                false ->
+                    case lists:member({Id, Rev}, ExcludeIdRevs) of
+                        true -> ok;
+                        false ->
+                            {ok, Doc} = couch_db:open_doc(Db, DI, []),
+                            Json = couch_doc:to_json_obj(Doc, []),
+                            [Fields|_] = proc_prompt(Proc, [<<"index_doc">>, Json]),
+                            Fields1 = [list_to_tuple(Field) || Field <- Fields],
+                            case Fields1 of
+                                [] ->
+                                    ok = clouseau_rpc:delete(IndexPid, Id);
+                                _  ->
+                                    ok = clouseau_rpc:update(IndexPid, Id, Fields1)
+                            end
+                    end
             end
     end,
     %% Force a commit every minute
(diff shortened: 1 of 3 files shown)


src/dreyfus_index.erl
```diff
@@ -228,9 +251,14 @@ open_index(DbName, #index{analyzer=Analyzer, sig=Sig}) ->
         {ok, Pid} ->
             case clouseau_rpc:get_update_seq(Pid) of
```
mayya-sharipova, line 229: it may be better to have a single message `get_update_purge_seqs` or smth, rather than to send two messages

src/dreyfus_index_updater.erl
```diff
@@ -17,20 +17,52 @@
...
+                    NewPurgeSeq = lists:max(PSeqs)
+                end;
+            true ->
+                %reset index
+                dreyfus_fabric_cleanup:go(DbName),
+                NewPurgeSeq = IdxPurgeSeq,
+                exit({reset, 0})
+            end
+        end,
+
```
mayya-sharipova: I would make lines 35-65 as a separate function, that either returns `{ok, NewPurgeSeq}` or `reset`.

src/dreyfus_index_updater.erl
```diff
@@ -17,20 +17,52 @@
...
+            FoldFun = fun(PurgeSeq, {Id, Revs}, Acc) ->
+                {ok, [ {PurgeSeq, Id, Revs} | Acc]}
+            end,
+            if (IdxPurgeSeq + 1) >= DbOldestPurgeSeq ->
+                {ok, PurgeSeqIdRevs} = couch_db:fold_purged_docs(Db, IdxPurgeSeq, FoldFun, [], []),
+
+                PSeqs = lists:foldr(fun({PSeq, Id, _}, Acc) ->
+                    clouseau_rpc:delete(IndexPid, Id),
+                    clouseau_rpc:set_purge_seq(IndexPid, PSeq),
```
mayya-sharipova: you don't need to do `set_purge_seq` in the loop. Similarly to update_seq you can commit it periodically (e.g. every minute) or at the end.

src/dreyfus_index_updater.erl
```diff
@@ -17,20 +17,52 @@
...
+                    [PSeq | Acc]
+                end, [], PurgeSeqIdRevs),
```
mayya-sharipova: Instead of list for Acc you can use just a starting purge_seq, so at the end you will get your final purge_seq for the accumulator, and you don't need to do `lists:max(PSeqs)` on line 56.

src/dreyfus_index_updater.erl
```diff
@@ -17,20 +17,52 @@
...
+
+                if PSeqs == [] ->
+                    NewPurgeSeq = IdxPurgeSeq;
+                true ->
+                    NewPurgeSeq = lists:max(PSeqs)
+                end;
+            true ->
+                %reset index
+                dreyfus_fabric_cleanup:go(DbName),
```
mayya-sharipova: This will deleted all search indexes for the database, we need just delete the current index.

src/dreyfus_index.erl
```diff
@@ -181,6 +181,29 @@ handle_info({'EXIT', FromPid, {updated, NewSeq}},
...
                              }}
     end;
+handle_info({'EXIT', Pid, {reset, NewSeq}},
+            #state{
+                updater_pid = Pid,
+                index_pid = IndexPid
+            } = State) ->
+    Index0 = State#state.index,
+    Index = Index0#index{current_seq = NewSeq},
+    NewSt = case reply_with_index(IndexPid ,Index, State#state.waiting_list) of
```
mayya-sharipova: we need to discuss in the #cloudant-search channel how to properly reset the index. I think if we are resetting the index, we need to kill the existing dreyfus_index process, corresponding clouseau_process, remove all corresponding lucene files, and after that initiate new clouseau & dreyfus processes. And this what dreyfus_index_manager should do

src/dreyfus_index_updater.erl
chguocloudant: I think the purge seq will be loss if an exception happened and the loop is not finished yet.

src/dreyfus_index_updater.erl
chguocloudant: You mean the we can get final_purge_seq by adding the count of `PurgeSeqIdRevs` and starting purge_seq?
is the `PurgeSeqIdRevs` consecutive?

src/dreyfus_index_updater.erl
mayya-sharipova: @chguocloudant  even with `clouseau_rpc:set_purge_seq` you are not committing it to disk, so if the process crashes, you will still loose a new `purge_seq`

src/dreyfus_index_updater.erl
mayya-sharipova: I mean this:
```
NewPurgeSeq = lists:foldr(fun({PSeq, Id, _}, _Acc) ->
                    clouseau_rpc:delete(IndexPid, Id),
                    clouseau_rpc:set_purge_seq(IndexPid, PSeq),
                    PSeq
                end, IdxPurgeSeq, PurgeSeqIdRevs),
```

chguocloudant: got it. 
And I will check if the `PurgeSeqIdRevs` is already sorted by PurgeSeq.

src/dreyfus_index_updater.erl
chguocloudant: got it, thank you.

src/dreyfus_index.erl
```diff
@@ -222,16 +210,32 @@ code_change(_OldVsn, State, _Extra) ->
...
     Path = <<DbName/binary,"/",Sig/binary>>,
     case clouseau_rpc:open_index(self(), Path, Analyzer) of
         {ok, Pid} ->
-            case clouseau_rpc:get_update_seq(Pid) of
-                {ok, Seq} ->
-                    {ok, Pid, Seq};
-                Error ->
-                    Error
-            end;
+            UpdateSeq = case clouseau_rpc:get_update_seq(Pid) of
```
rnewson: why did this section change?

src/dreyfus_index.erl
```diff
@@ -164,23 +164,11 @@ handle_info({'EXIT', FromPid, {updated, NewSeq}},
             #state{
               index=Index0,
               index_pid=IndexPid,
-              updater_pid=UpPid,
-              waiting_list=WaitList
+              updater_pid=UpPid
              }=State) when UpPid == FromPid ->
     Index = Index0#index{current_seq=NewSeq},
-    case reply_with_index(IndexPid, Index, WaitList) of
```
rnewson, line 171: why did this section change?

src/dreyfus_index.erl
```diff
@@ -222,16 +210,32 @@ code_change(_OldVsn, State, _Extra) ->
 
 % private functions.
 
+reply_or_update(IndexPid, Index, State) ->
```
rnewson: why was this function extracted?

src/dreyfus_index_updater.erl
```diff
@@ -64,7 +73,7 @@ update(IndexPid, Index) ->
         couch_db:close(Db)
     end.
 
-load_docs(FDI, _, {I, IndexPid, Db, Proc, Total, LastCommitTime}=Acc) ->
+load_docs(FDI, {I, IndexPid, Db, Proc, Total, LastCommitTime}=Acc) ->
```
rnewson: why did this argument disappear?

rnewson (commented):
This PR mixes the purge changes along with stylistic changes and refactorings. Please remove everything except what is needed to support index purge.

src/dreyfus_index.erl
chguocloudant: There are some changes here before, but they are removed.
I will change it back to origin status.

src/dreyfus_index.erl
chguocloudant: these lines are extracted to a new function to make it easier to understand.
I followed the comments in hastings from Paul Davis. But the new function here is only called once.

src/dreyfus_index_updater.erl
chguocloudant: We used a new function `couch_db:fold_changes` here, and this function need a function with 2 arguments.

src/dreyfus_index.erl
chguocloudant: I have modified it and commit it later.

src/dreyfus_index_updater.erl
```diff
@@ -31,6 +31,14 @@ update(IndexPid, Index) ->
     erlang:put(io_priority, {view_update, DbName, IndexName}),
     {ok, Db} = couch_db:open_int(DbName, []),
     try
+        {ok, IdxPurgeSeq} = clouseau_rpc:get_purge_seq(IndexPid),
+        FoldFun = fun(PurgeSeq, {Id, _Revs}, Acc) ->
+            clouseau_rpc:delete(IndexPid, Id),
```
rnewson: I don't think this is the correct behaviour. The purge API is on id/rev pairs, the request might leave some revisions alone, in which case the index should reflect that view. This code appears to remove the document from the index in all cases.

src/dreyfus_index_updater.erl
rnewson: Why did the call to enum_docs_since change to a call to fold_changes?

src/dreyfus_rpc.erl
```diff
@@ -47,29 +47,31 @@ call(Fun, DbName, DDoc, IndexName, QueryArgs0) ->
...
+                    Result = dreyfus_index:Fun(Pid, QueryArgs),
+                    rexi:reply(Result);
                 Error ->
                     rexi:reply(Error)
             end;
         Error ->
             rexi:reply(Error)
     end.
 
+get_index_pid(DbName, Index) ->
```
rnewson: Please remove the extraction of get_index_pid from this PR so we can focus on the purge feature change.

test/dreyfus_purge_test.erl
```diff
@@ -0,0 +1,186 @@
...
+-module(dreyfus_purge_test).
+
+-include_lib("couch/include/couch_db.hrl").
+-include_lib("dreyfus/include/dreyfus.hrl").
+
+-export([test_purge_single/0, test_purge_multiple/0, test_purge_multiple2/0, test_all/0]).
+-export([create_db_docs/1, create_docs/2, delete_db/1, purge_one_doc/2, dreyfus_search/2]).
+
+test_all() ->
+    test_purge_single(),
```
rnewson, line 26: It would help if you added a test where only some revisions of a document are purged, we can then assert that the remaining revision is indexed correctly.

rnewson (changes requested):

src/dreyfus_rpc.erl
chguocloudant: OK, I will remove this function.

test/dreyfus_purge_test.erl
chguocloudant: OK, I will make a new function to test this case.

src/dreyfus_index_updater.erl
chguocloudant: This change is used for pluggable storage engine API. The purge issue is based on plugable storage engine API.
But I think it should be commit separately.

src/dreyfus_index_updater.erl
chguocloudant: This `FoldFun` is enum the purged docs, and delete the invalid index. But the index will be rebuilt later.
The purge API is work on id/rev pairs, but one doc has only one index. So if the doc is changed, the index should be delete first and then rebuild it.

rnewson:
I'm out on vacation until Jan 3rd but can comment today. 

That's not how we want to update the index. It should go from its current state to the new state without an intermediate state where the document is deleted from the index.

chguocloudant:
Hi Robert, good day.
 
I think I should express our steps in purging index.This is our ways to purge and rebuild index as below.
 
If a doc has only one leaf node. This doc is purged and also be deleted.Search this doc, and trigged the function `dreyfus_index_updater:update`.Firstly, fold purged docs by `couch_db:fold_purged_docs/5`. And the doc's index will be deleted in the fold function.Secondly, fold updated docs by `couch_db:fold_changes` and the fold function is `load_docs/2`. The doc is deleted already, so do nothing about this doc in the fold function.
 
If a doc has more then one leaf node. One of the leaf node is purged but this doc is not deleted.Search this doc, and trigged the function `dreyfus_index_updater:update`.Firstly, fold purged docs by `couch_db:fold_purged_docs/5`. And the doc's index will be deleted in the fold function.Secondly, fold updated docs by `couch_db:fold_changes` and the fold function is `load_docs/2`. The doc exist but there is no index, so the index will be updated(rebuilt).
 
I think it achieves the result we want too.
 
I understand your meaning is that the doc should goes a)from original state to deleted state or b)from original state to updated state in fold function when we fold the purge seq. Just like the operations in `load_docs/2`.a) the doc is purged and deleted.b) the doc is purged but not deleted.And current code goes from original state to deleted state first and then goes to updated state if the doc still exist.
 
But we can only get `Id` and purged `Rev` in the fold function, so we don't have enough info to update(rebuild) index directly here.
 
What do you think?
 
Chang Hua Guo (郭昌华)
E-mail: chguo@cn.ibm.com
 
 
----- Original message -----From: Robert Newson <notifications@github.com>To: cloudant-labs/dreyfus <dreyfus@noreply.github.com>Cc: Chang Hua DG Guo/China/IBM@IBMCN, Mention <mention@noreply.github.com>Subject: Re: [cloudant-labs/dreyfus] Support index purge in dreyfus. (#11)Date: Fri, Dec 23, 2016 4:58 PM I'm out on vacation until Jan 3rd but can comment today.That's not how we want to update the index. It should go from its current state to the new state without an intermediate state where the document is deleted from the index.Sent from my iPhone> On 23 Dec 2016, at 07:47, Chang Hua Guo <notifications@github.com> wrote:>> @chguocloudant commented on this pull request.>> In src/dreyfus_index_updater.erl:>> > @@ -31,6 +31,14 @@ update(IndexPid, Index) ->> erlang:put(io_priority, {view_update, DbName, IndexName}),> {ok, Db} = couch_db:open_int(DbName, []),> try> + {ok, IdxPurgeSeq} = clouseau_rpc:get_purge_seq(IndexPid),> + FoldFun = fun(PurgeSeq, {Id, _Revs}, Acc) ->> + clouseau_rpc:delete(IndexPid, Id),> This FoldFun is enum the purged docs, and delete the invalid index. But the index will be rebuilt later.> The purge API is work on id/rev pairs, but one doc has only one index. So if the doc is changed, the index should be delete first and then rebuild it.>> —> You are receiving this because you commented.> Reply to this email directly, view it on GitHub, or mute the thread.>

src/dreyfus_index_updater.erl
rnewson: ok, we can't merge this PR until the pluggable storage engine API is also merged.

src/dreyfus_index_updater.erl
rnewson: Hm, I'm still not convinced. If we look at couch_index/couch_mrview that seems to do the right thing. It purges the items and then updates the index, before returning the updated mrview state to the caller. Here we will make the intermediate state, where the Documents in the Lucene index are removed, visible to search results.

chguocloudant:
Hi @rnewson . 
I am not sure if I express myself clear enough. If you have any question please leave some comments.
Thank you very much.

src/dreyfus_index_updater.erl
chguocloudant: I read couch_index/couch_mrview code and I think it is same to drefus. It is also delete first and then rebuild it.
If we purged a document and then search doc. Here is the steps in dreyfus.
1. fold purged docs and it will delete the indexes of purged docs.
2. fold db changes and it will rebuild the indexes of docs if they are still exist.
3. do search.

The operations are triggered by search request. Delete and update index first and then search and return the result to the caller.

src/dreyfus_index_updater.erl
```diff
@@ -89,3 +101,43 @@ load_docs(FDI, _, {I, IndexPid, Db, Proc, Total, LastCommitTime}=Acc) ->
         false ->
             {ok, setelement(1, Acc, I+1)}
     end.
+
+purge_index(Db, IndexPid, Index) ->
+    {ok, IdxPurgeSeq} = clouseau_rpc:get_purge_seq(IndexPid),
+    Proc = get_os_process(Index#index.def_lang),
```
rnewson, line 88: Proc needs to be returned when we're done (a try/after is usual).

src/dreyfus_index_updater.erl
```diff
@@ -89,3 +101,43 @@ load_docs(FDI, _, {I, IndexPid, Db, Proc, Total, LastCommitTime}=Acc) ->
...
+                                Fields1 = [list_to_tuple(Field) || Field <- Fields],
+                                case Fields1 of
+                                    [] ->
+                                        ok = clouseau_rpc:delete(IndexPid, Id);
+                                    _  ->
+                                        ok = clouseau_rpc:update(IndexPid, Id, Fields1)
+                                end
+                        end,
+                        lists:keydelete(Id, 1, Acc),
+                        Acc0 = lists:append([{Id, Rev}], Acc)
```
rnewson: `[{Id, Rev} | Acc]` is more idiomatic (and faster). If order matters, then reverse the list at the end.

src/dreyfus_index_updater.erl
```diff
@@ -89,3 +101,43 @@ load_docs(FDI, _, {I, IndexPid, Db, Proc, Total, LastCommitTime}=Acc) ->
...
+
+purge_index(Db, IndexPid, Index) ->
+    {ok, IdxPurgeSeq} = clouseau_rpc:get_purge_seq(IndexPid),
+    Proc = get_os_process(Index#index.def_lang),
+    true = proc_prompt(Proc, [<<"add_fun">>, Index#index.def]),
+    FoldFun = fun(PurgeSeq, {Id, _Revs}, Acc) ->
+        case couch_db:get_full_doc_info(Db, Id) of
+            not_found ->
+                ok = clouseau_rpc:delete(IndexPid, Id),
+                Acc0 = Acc;
```
rnewson: just `Acc;` here. Rather than assign a variable in every path, it's more idiomatic to use the result of the case statement (i.e, `Acc0 = case couch_db...`).

src/dreyfus_index_updater.erl
```diff
@@ -89,3 +101,43 @@ load_docs(FDI, _, {I, IndexPid, Db, Proc, Total, LastCommitTime}=Acc) ->
...
+                        Acc0 = Acc;
+                    false ->
+                        case Del of
+                            true ->
+                                ok = clouseau_rpc:delete(IndexPid, Id);
+                            false ->
+                                {ok, Doc} = couch_db:open_doc(Db, DI, []),
+                                Json = couch_doc:to_json_obj(Doc, []),
+                                [Fields|_] = proc_prompt(Proc, [<<"index_doc">>, Json]),
+                                Fields1 = [list_to_tuple(Field) || Field <- Fields],
```
rnewson: this section is a cut-and-paste from elsewhere in the file, please extract it as a function that is called from both places instead.

src/dreyfus_index_updater.erl
```diff
@@ -89,3 +101,43 @@ load_docs(FDI, _, {I, IndexPid, Db, Proc, Total, LastCommitTime}=Acc) ->
...
+                                case Fields1 of
+                                    [] ->
+                                        ok = clouseau_rpc:delete(IndexPid, Id);
+                                    _  ->
+                                        ok = clouseau_rpc:update(IndexPid, Id, Fields1)
+                                end
+                        end,
+                        lists:keydelete(Id, 1, Acc),
```
rnewson: this does nothing as the result is discarded.

rnewson (changes requested):

src/dreyfus_index_updater.erl
chguocloudant: Got it. I will modify it and commit it later.

src/dreyfus_index_updater.erl
chguocloudant: Got it, will modify and commit it later.

src/dreyfus_index_updater.erl
chguocloudant: OK, I will extract it as a new function.

src/dreyfus_index_updater.erl
chguocloudant: Sorry, I will remove it.

src/dreyfus_index_updater.erl
chguocloudant: Got it. will modify it.

test/dreyfus_purge_test.erl
```diff
@@ -0,0 +1,543 @@
...
+    Query = <<"color:red">>,
+
+    %create the db and docs
+    DbName = db_name(),
+    create_db_docs(DbName),
+
+    %first search request
+    {ok, _, HitCount1, _, _, _} = dreyfus_search(DbName, Query),
+
+    ?assertEqual(HitCount1, 5),
```
rnewson, line 60: this could be improved with `?assertMatch`, i.e;

`?assertMatch({ok, _, 5, _, _, _}, dreyfus_search(DbName, Query)),`

test/dreyfus_purge_test.erl
```diff
@@ -0,0 +1,543 @@
...
+    %get full dbname
+    DBFullName = get_full_dbname(DbName),
+
+    %purge 5 docs
+    purge_docs(DBFullName, [<<"apple">>, <<"tomato">>, <<"cherry">>, <<"haw">>, <<"strawberry">>]),
+
+    %second search request
+    {ok, _, HitCount2, _, _, _} = dreyfus_search(DbName, Query),
+
+    ?assertEqual(HitCount2, 0),
```
rnewson, line 71: same comment as above

rnewson (approved):
I'd like to see the eunit tests improved as indicated but I think this work is complete, thanks!

A reminder that we can't merge this until the pluggable storage API is merged into couchdb master.

Also, we might have a concern with the idrev list we build during the fold which helps us not make duplicate updates to the search index. If that was a very large list, it might be a problem. What do you think?

test/dreyfus_purge_test.erl
chguocloudant: Thanks, got it, I will modify it.

test/dreyfus_purge_test.erl
chguocloudant: Thanks, got it, I will modify it.

chguocloudant:
Thanks. I won't merge it until the pluggable storage API is merged into couchdb master.
The size of the IdRev list depends on how many docs user purged in a period. I think the size won't be too large generally.
But if this happened, I don't think what can be done in dreyfus to fix this problem. We can try to use some more effective functions to shorten execution time. 
I think we can set a threshold to trigger the index purging automatically. It maybe can avoid this problem.

PR #12, opened 2017-02-10
81816 clustered purge improvements

chguocloudant:
Update for Clustered Purge Improvement

 - add verify_index_exists function to check if the index exist.
 - update utc_timestamp function.
 - add some test cases.

BugzId: 81816

### DIFF
--- a/src/dreyfus_fabric_cleanup.erl
+++ b/src/dreyfus_fabric_cleanup.erl
@@ -18,7 +18,6 @@
 -include("dreyfus.hrl").
 -include_lib("mem3/include/mem3.hrl").
 -include_lib("couch/include/couch_db.hrl").
--include_lib("couch_mrview/include/couch_mrview.hrl").
 
 -export([go/1]).
 
@@ -57,8 +56,13 @@ cleanup_local_purge_doc(DbName, ActiveSigs) ->
     DeleteDocs = lists:map(fun(IdxDir) ->
         IdxDirList = filename:split(IdxDir),
         [Sig] = lists:nthtail(length(IdxDirList)-1, IdxDirList),
-        DocId = dreyfus_util:get_local_purge_doc_id(Sig),
-        #doc{id = DocId, deleted=true}
+        case re:run(Sig, "^[a-fA-F0-9]+$" ,[{capture, none}]) of
+            match ->
+                DocId = dreyfus_util:get_local_purge_doc_id(Sig),
+                #doc{id = DocId, deleted=true};
+            _ ->
+                []
+        end
     end, DeadDirs),
-    fabric:update_docs(DbName, DeleteDocs, [?ADMIN_CTX]),
+    fabric:update_docs(DbName, lists:flatten(DeleteDocs), [?ADMIN_CTX]),
     ok.
--- a/src/dreyfus_fabric_cleanup.erl
+++ b/src/dreyfus_fabric_cleanup.erl
@@ -18,6 +18,7 @@
 -include("dreyfus.hrl").
 -include_lib("mem3/include/mem3.hrl").
 -include_lib("couch/include/couch_db.hrl").
+-include_lib("couch_mrview/include/couch_mrview.hrl").
 
 -export([go/1]).
 
@@ -25,6 +26,7 @@ go(DbName) ->
     {ok, DesignDocs} = fabric:design_docs(DbName),
     ActiveSigs = lists:usort(lists:flatmap(fun active_sigs/1,
         [couch_doc:from_json_obj(DD) || DD <- DesignDocs])),
+    cleanup_local_purge_doc(DbName, ActiveSigs),
     clouseau_rpc:cleanup(DbName, ActiveSigs),
     ok.
 
@@ -35,3 +37,28 @@ active_sigs(#doc{body={Fields}}=Doc) ->
          {ok, Index} = dreyfus_index:design_doc_to_index(Doc, IndexName),
          Index#index.sig
      end || IndexName <- IndexNames].
+
+cleanup_local_purge_doc(DbName, ActiveSigs) ->
+    {ok, BaseDir} = clouseau_rpc:get_root_dir(),
+    DbNamePattern = <<DbName/binary, ".*">>,
+    Pattern0 = filename:join([BaseDir, "shards", "*", DbNamePattern, "*"]),
+    Pattern = binary_to_list(iolist_to_binary(Pattern0)),
+    DirListStrs = filelib:wildcard(Pattern),
+    DirList = [iolist_to_binary(DL) || DL <- DirListStrs],
+    LocalShards = mem3:local_shards(DbName),
+    ActiveDirs = lists:foldl(fun(LS, AccOuter) ->
+        lists:foldl(fun(Sig, AccInner) ->
+            DirName = filename:join([BaseDir, LS#shard.name, Sig]),
+            [DirName | AccInner]
+        end, AccOuter, ActiveSigs)
+    end, [], LocalShards),
+
+    DeadDirs = DirList -- ActiveDirs,
+    DeleteDocs = lists:map(fun(IdxDir) ->
+        IdxDirList = filename:split(IdxDir),
+        [Sig] = lists:nthtail(length(IdxDirList)-1, IdxDirList),
+        DocId = dreyfus_util:get_local_purge_doc_id(Sig),
+        #doc{id = DocId, deleted=true}
+    end, DeadDirs),
+    fabric:update_docs(DbName, DeleteDocs, [?ADMIN_CTX]),
+    ok.
--- a/src/dreyfus_index.erl
+++ b/src/dreyfus_index.erl
@@ -385,21 +385,15 @@ info_int(Pid) ->
     clouseau_rpc:info(Pid).
 
 verify_index_exists(Options) ->
-    DbName = dreyfus_util:get_value_from_options(<<"db_name">>, Options),
+    DbName0 = dreyfus_util:get_value_from_options(<<"dbname">>, Options),
     DDocId = dreyfus_util:get_value_from_options(<<"ddoc_id">>, Options),
-    IdxName = dreyfus_util:get_value_from_options(<<"index_name">>, Options),
-    Sig = dreyfus_util:get_value_from_options(<<"sig">>, Options),
-    case couch_db:open_int(DbName, []) of
-        {ok, Db} ->
-            try
-                {ok, DDoc} = couch_db:open_doc(Db, DDocId, []),
-                {ok, Index} = dreyfus_index:design_doc_to_index(DDoc, IdxName),
-                Sig == Index#index.sig
-            catch _:_ ->
-                false
-            after
-                couch_db:close(Db)
-            end;
-        _ ->
-            false
+    IdxName = dreyfus_util:get_value_from_options(<<"indexname">>, Options),
+    Sig = dreyfus_util:get_value_from_options(<<"signature">>, Options),
+    try
+        DbName = mem3:dbname(DbName0),
+        {ok, DDoc} = ddoc_cache:open(DbName, DDocId),
+        {ok, Index} = dreyfus_index:design_doc_to_index(DDoc, IdxName),
+        Sig == Index#index.sig
+    catch _:_ ->
+        false
     end.
--- a/src/dreyfus_index.erl
+++ b/src/dreyfus_index.erl
@@ -25,7 +25,7 @@
 % public api.
 -export([start_link/2, design_doc_to_index/2, await/2, search/2, info/1,
          group1/2, group2/2,
-         design_doc_to_indexes/1]).
+         design_doc_to_indexes/1, verify_index_exists/1]).
 
 % gen_server api.
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2,
@@ -109,6 +109,7 @@ init({DbName, Index}) ->
             case couch_db:open_int(DbName, []) of
                 {ok, Db} ->
                     try couch_db:monitor(Db) after couch_db:close(Db) end,
+                    maybe_create_local_purge_doc(Db, Pid, Index),
                     proc_lib:init_ack({ok, self()}),
                     gen_server:enter_loop(?MODULE, [], State);
                 Error ->
@@ -116,6 +117,19 @@ init({DbName, Index}) ->
             end;
         Error ->
             proc_lib:init_ack(Error)
+    end.
+
+maybe_create_local_purge_doc(Db, IndexPid, Index) ->
+    DocId = dreyfus_util:get_local_purge_doc_id(Index#index.sig),
+    case couch_db:open_doc(Db, DocId) of
+        {not_found, _} ->
+            DbPurgeSeq = couch_db:get_purge_seq(Db),
+            clouseau_rpc:set_purge_seq(IndexPid, DbPurgeSeq),
+            DocContent = dreyfus_util:get_local_purge_doc_body(
+                Db, DocId, DbPurgeSeq, Index),
+            couch_db:update_doc(Db, DocContent, []);
+        _ ->
+            ok
     end.
 
 handle_call({await, RequestSeq}, From,
@@ -369,3 +383,23 @@ group2_int(Pid, QueryArgs0) ->
 
 info_int(Pid) ->
     clouseau_rpc:info(Pid).
+
+verify_index_exists(Options) ->
+    DbName = dreyfus_util:get_value_from_options(<<"db_name">>, Options),
+    DDocId = dreyfus_util:get_value_from_options(<<"ddoc_id">>, Options),
+    IdxName = dreyfus_util:get_value_from_options(<<"index_name">>, Options),
+    Sig = dreyfus_util:get_value_from_options(<<"sig">>, Options),
+    case couch_db:open_int(DbName, []) of
+        {ok, Db} ->
+            try
+                {ok, DDoc} = couch_db:open_doc(Db, DDocId, []),
+                {ok, Index} = dreyfus_index:design_doc_to_index(DDoc, IdxName),
+                Sig == Index#index.sig
+            catch _:_ ->
+                false
+            after
+                couch_db:close(Db)
+            end;
+        _ ->
+            false
+    end.
--- a/src/dreyfus_index_updater.erl
+++ b/src/dreyfus_index_updater.erl
@@ -69,29 +69,10 @@ update(IndexPid, Index) ->
 load_docs(FDI, {I, IndexPid, Db, Proc, Total, LastCommitTime, ExcludeIdRevs}=Acc) ->
     couch_task_status:update([{changes_done, I}, {progress, (I * 100) div Total}]),
     DI = couch_doc:to_doc_info(FDI),
-    #doc_info{id=Id, high_seq=Seq, revs=[#rev_info{rev=Rev, deleted=Del}|_]} = DI,
+    #doc_info{id=Id, high_seq=Seq, revs=[#rev_info{rev=Rev}|_]} = DI,
     case lists:member({Id, Rev}, ExcludeIdRevs) of
         true -> ok;
-        false ->
-            case Del of
-                true ->
-                    ok = clouseau_rpc:delete(IndexPid, Id);
-                false ->
-                    case lists:member({Id, Rev}, ExcludeIdRevs) of
-                        true -> ok;
-                        false ->
-                            {ok, Doc} = couch_db:open_doc(Db, DI, []),
-                            Json = couch_doc:to_json_obj(Doc, []),
-                            [Fields|_] = proc_prompt(Proc, [<<"index_doc">>, Json]),
-                            Fields1 = [list_to_tuple(Field) || Field <- Fields],
-                            case Fields1 of
-                                [] ->
-                                    ok = clouseau_rpc:delete(IndexPid, Id);
-                                _  ->
-                                    ok = clouseau_rpc:update(IndexPid, Id, Fields1)
-                            end
-                    end
-            end
+        false -> update_or_delete_index(IndexPid, Db, DI, Proc)
     end,
     %% Force a commit every minute
     case timer:now_diff(Now = now(), LastCommitTime) >= 60000000 of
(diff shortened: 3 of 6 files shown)


src/dreyfus_fabric_cleanup.erl
```diff
@@ -26,10 +26,20 @@ go(DbName) ->
     {ok, DesignDocs} = fabric:design_docs(DbName),
     ActiveSigs = lists:usort(lists:flatmap(fun active_sigs/1,
         [couch_doc:from_json_obj(DD) || DD <- DesignDocs])),
+    cleanup_local_doc(DbName, ActiveSigs),
     clouseau_rpc:cleanup(DbName, ActiveSigs),
+    ok.
+
+active_sigs(#doc{body={Fields}}=Doc) ->
```
mayya-sharipova: Why `active_sigs` has been moved? Looks like a new code.

src/dreyfus_fabric_cleanup.erl
```diff
@@ -26,10 +26,20 @@ go(DbName) ->
...
+    {IndexNames, _} = lists:unzip(RawIndexes),
+    [begin
+         {ok, Index} = dreyfus_index:design_doc_to_index(Doc, IndexName),
+         Index#index.sig
+     end || IndexName <- IndexNames].
 
-    %clean local doc
-    BaseDir = config:get("couchdb", "search_index_dir", "/srv/search_index"),
+cleanup_local_doc(DbName, ActiveSigs) ->
+    {ok, BaseDir} = clouseau_rpc:get_root_dir(),
```
mayya-sharipova: What is the reason that you are not using config to get `BaseDir`?

src/dreyfus_fabric_cleanup.erl
```diff
@@ -35,3 +37,41 @@ active_sigs(#doc{body={Fields}}=Doc) ->
          {ok, Index} = dreyfus_index:design_doc_to_index(Doc, IndexName),
          Index#index.sig
      end || IndexName <- IndexNames].
+
+cleanup_local_doc(DbName, ActiveSigs) ->
```
mayya-sharipova: I am not sure if we need this complex code to delete _local/purge-search ... docs. Why do we need to open all  8 shards dbs and do it in a loop?
I would just use the following code:

```
cleanup_local_doc(DbName, ActiveSigs) ->
   Opts = [{start_key, list_to_binary(?LOCAL_DOC_PREFIX ++ "purge-search-")},
        {end_key_gt, list_to_binary(?LOCAL_DOC_PREFIX ++ "purge-search1")}],
       % use fabric all_docs with Ops to get  all _local/purge-search docs
       {ok, LocalPDocs} = fabric:all_docs(DbName,....)
        % filter those the need to be deleted
        % use fabric update_docs for docs that need to be removed
        fabric:update_docs(

``

File: src/dreyfus_fabric_cleanup.erl (hunk 2 above)
Comment (chguocloudant) [reply src/dreyfus_fabric_cleanup.erl]: Hi Mayya. 
a) I checked the config file and there is no record of `search_index_dir`.
b) The search index dir is controlled by `clouseau` in fact, so I think it is better to get `BaseDir` from `clouseau`.

File: src/dreyfus_fabric_cleanup.erl (hunk 1 above)
Comment (chguocloudant) [reply src/dreyfus_fabric_cleanup.erl]: Hi Mayya. I didn't change this function. I think it is a mistake by compare tool.

File: src/dreyfus_fabric_cleanup.erl (hunk 3 above)
Comment (chguocloudant) [reply src/dreyfus_fabric_cleanup.erl]: Hi Mayya. The function `fabric:all_docs` cannot collect all the local docs because there is an error in function `fold_local_docs` which is used.

File: src/dreyfus_fabric_cleanup.erl [hunk 4]
```diff
@@ -18,13 +18,15 @@
...
 -include_lib("couch/include/couch_db.hrl").
+-include_lib("couch_mrview/include/couch_mrview.hrl").
 
 -export([go/1]).
 
 go(DbName) ->
     {ok, DesignDocs} = fabric:design_docs(DbName),
     ActiveSigs = lists:usort(lists:flatmap(fun active_sigs/1,
         [couch_doc:from_json_obj(DD) || DD <- DesignDocs])),
+    cleanup_local_doc(DbName, ActiveSigs),
```
mayya-sharipova: Would `cleanup_local_purge_docs` be a better name?

chguocloudant: yes, you are right. I will modify it later. Thank you.

src/dreyfus_fabric_cleanup.erl
```diff
@@ -35,3 +37,28 @@ active_sigs(#doc{body={Fields}}=Doc) ->
...
+
+    DeadDirs = DirList -- ActiveDirs,
+    DeleteDocs = lists:map(fun(IdxDir) ->
+        IdxDirList = filename:split(IdxDir),
+        [Sig] = lists:nthtail(length(IdxDirList)-1, IdxDirList),
+        DocId = dreyfus_util:get_local_purge_doc_id(Sig),
+        #doc{id = DocId, deleted=true}
+    end, DeadDirs),
+    fabric:update_docs(DbName, DeleteDocs, [?ADMIN_CTX]),
+    ok.
```
mayya-sharipova: LGTM (looks good to me)!

src/dreyfus_index.erl
```diff
@@ -118,6 +119,16 @@ init({DbName, Index}) ->
             proc_lib:init_ack(Error)
     end.
 
+maybe_create_local_purge_doc(Db, Index) ->
+    DocId = dreyfus_util:get_local_purge_doc_id(Index#index.sig),
+    case couch_db:open_doc(Db, DocId) of
+        {not_found, _} ->
+            DocContent = dreyfus_util:get_local_purge_doc_body(Db, DocId, 0, Index),
```
mayya-sharipova: @chguocloudant  I think, Initial purge_seq that you pass to `dreyfus_util:get_local_purge_doc_body` should be NOT 0, but the database's purge_seq.

src/dreyfus_index_updater.erl
```diff
@@ -64,10 +66,55 @@ update(IndexPid, Index) ->
...
+                            update_or_delete_index(IndexPid, Db, DI, Proc),
+                            update_local_doc(Db, Index, PurgeSeq),
+                            [{Id, Rev} | Acc]
+                    end
+            end,
+            clouseau_rpc:set_purge_seq(IndexPid, PurgeSeq),
+            {ok, Acc0}
+        end,
+        couch_db:fold_purged_docs(Db, IdxPurgeSeq, FoldFun, [], [])
+    after
```
mayya-sharipova: @chguocloudant  A new format of `fold_purged_docs` function is following:
```
FoldFun = fun({PurgeSeq ,_UUID, Id, Revs}, Acc) ->  ... NewAcc end,
{ok, PurgedIdsRevs} = couch_db:fold_purged_docs(Db, IdxPurgeSeq, FoldFun, [], []),
```

src/dreyfus_index_updater.erl
```diff
@@ -64,10 +66,55 @@ update(IndexPid, Index) ->
...
+purge_index(Db, IndexPid, Index) ->
+    {ok, IdxPurgeSeq} = clouseau_rpc:get_purge_seq(IndexPid),
+    Proc = get_os_process(Index#index.def_lang),
+    try
+        true = proc_prompt(Proc, [<<"add_fun">>, Index#index.def]),
+        FoldFun = fun(PurgeSeq, {Id, _Revs}, Acc) ->
+            Acc0 = case couch_db:get_full_doc_info(Db, Id) of
+                not_found ->
+                    ok = clouseau_rpc:delete(IndexPid, Id),
+                    update_local_doc(Db, Index, PurgeSeq),
```
mayya-sharipova: @chguocloudant  Would a name `update_local_purge_doc` be better? 

Do you think we can do `update_local_doc` and `clouseau_rpc:set_purge_seq(IndexPid, PurgeSeq)` outside of the loop with the final purge_seq?

src/dreyfus_index_updater.erl
```diff
@@ -64,10 +66,55 @@ update(IndexPid, Index) ->
...
+                    Acc;
+                FDI ->
+                    DI = couch_doc:to_doc_info(FDI),
+                    #doc_info{id=Id, revs=[#rev_info{rev=Rev}|_]} = DI,
+                    case lists:member({Id, Rev}, Acc) of
```
mayya-sharipova, line 102: It is not clear what you are checking here  with this `case` statement - can you put a comment for this

src/dreyfus_index_updater.erl
```diff
@@ -64,10 +66,55 @@ update(IndexPid, Index) ->
...
+                        true -> Acc;
+                        false ->
+                            update_or_delete_index(IndexPid, Db, DI, Proc),
```
mayya-sharipova, line 105: would instead of `update_or_delete_index` a better name be `update_or_delete_doc`?

src/dreyfus_index_updater.erl
```diff
@@ -31,6 +31,7 @@ update(IndexPid, Index) ->
     erlang:put(io_priority, {view_update, DbName, IndexName}),
     {ok, Db} = couch_db:open_int(DbName, []),
     try
+        {ok, ExcludeIdRevs} = purge_index(Db, IndexPid, Index),
```
mayya-sharipova: @chguocloudant What is the reason you are collecting this `ExcludeIdRevs`? Would a simplier approach be just do `clouseau_rpc:delete(IndexPid, Id);` for all docs in `purge_index` function, and then keep `load_docs` function as it was before without any check `case lists:member({Id, Rev}, ExcludeIdRevs) `.

With my suggestion, If a doc is fully purged, it will be deleted in `purge_index` function using `clouseau_rpc:delete`. If a doc is partially purged, we will first also delete it in `purge_index`, and then the doc's new revision should come through changes to the `load_docs` function, and should be update there.

chguocloudant: Hi @mayya-sharipova . If one doc is purged, there will be a purge seq and db update seq. The index delete/update will be run twice. So we collect the `ExcludeIdRevs` in `purge_index` to avoid this situation.
This change is in another PR(<https://github.com/cloudant-labs/dreyfus/pull/11>). 
The new PR(#12) I created is based on it and the commit history is kept.

src/dreyfus_index_updater.erl
chguocloudant: I think `update_or_delete_index` will be OK. It might be ambiguous if it is named as a doc.

src/dreyfus_index_updater.erl
chguocloudant: If a doc has more Revs and purged twice or more, there will be twice or more purge seq for this doc. So the index will be update/delete twice or more. In fact it will be OK only once.
So I check the `{Id, Rev}` to keep update/delete once.
This change is in another PR(#11).
The new PR(#12) I created is based on it and the commit history is kept.

src/dreyfus_index_updater.erl
chguocloudant: Thank you Mayya, will modify it later.

src/dreyfus_index_updater.erl
chguocloudant: Hi @mayya-sharipova . I want to confirm that what is the order of the purge seq.
Should I check and compare to get the largest purge seq?
Or the first/last seq is the final(largest) purge seq?

src/dreyfus_index.erl
chguocloudant: I think if the local doc is not exist the purge seq is 0.
But I will modify it and the code will be more strong.
Thank you.

src/dreyfus_index.erl
```diff
@@ -119,11 +119,12 @@ init({DbName, Index}) ->
...
 
-maybe_create_local_purge_doc(Db, Index) ->
+maybe_create_local_purge_doc(Db, IndexPid, Index) ->
     DocId = dreyfus_util:get_local_purge_doc_id(Index#index.sig),
+    {ok, PurgeSeq} = clouseau_rpc:get_purge_seq(IndexPid),
     case couch_db:open_doc(Db, DocId) of
         {not_found, _} ->
-            DocContent = dreyfus_util:get_local_purge_doc_body(Db, DocId, 0, Index),
+            DocContent = dreyfus_util:get_local_purge_doc_body(Db, DocId, PurgeSeq, Index),
             couch_db:update_doc(Db, DocContent, []);
```
mayya-sharipova: @chguocloudant We need to think here, if the  initial purge_seq should be the **database's** purge_seq (or from clouseau).  

Consider this, you are having `maybe_create_local_purge_doc` function for the  upgrade. Before this upgrade `clouseau` was not storing any purge_seq (that what makes this index different from hastings).  So, initially you would like to get purge_seq from a database. 

You can put line #124 after line #126, so to request purge_seq only when necessary.

src/dreyfus_index.erl
```diff
@@ -388,18 +389,18 @@ verify_index_exists(Options) ->
...
-    case couch_db:open_doc(Db, DDocId, []) of
-        {ok, DDoc} ->
-            {ok, #index{sig=Sig0}} = dreyfus_index:design_doc_to_index(DDoc, IdxName),
-            Sig == Sig0;
+            try
+                {ok, DDoc} = couch_db:open_doc(Db, DDocId, []),
+                {ok, #index{sig=Sig0}} = dreyfus_index:design_doc_to_index(DDoc, IdxName),
+                Sig == Sig0
+            catch _:_ ->
+                couch_log:warning("Something wrong happened, please check it. "
```
mayya-sharipova: I think instead of `couch_log:warning` returning false is enough.  Also, a log contains thousands of messages, and when writing smth to log, you should specify the content what is the problem, and where it is coming from; otherwise it would be very difficult to debut it.

src/dreyfus_index.erl
```diff
@@ -118,6 +119,17 @@ init({DbName, Index}) ->
             proc_lib:init_ack(Error)
     end.
 
+maybe_create_local_purge_doc(Db, IndexPid, Index) ->
+    DocId = dreyfus_util:get_local_purge_doc_id(Index#index.sig),
+    {ok, PurgeSeq} = clouseau_rpc:get_purge_seq(IndexPid),
```
mayya-sharipova: @chguocloudant We need to think here, if the initial purge_seq should be the database's purge_seq (or from clouseau).

Consider this, you are having maybe_create_local_purge_doc function for the upgrade. Before this upgrade clouseau was not storing any purge_seq (that what makes this index different from hastings). So, initially you would like to get purge_seq from a database.

Or if you have different plan for upgrade, please suggest it.

You can put line #124 after line #126, so to request purge_seq only when necessary.

src/dreyfus_index_updater.erl
```diff
@@ -101,14 +100,16 @@ purge_index(Db, IndexPid, Index) ->
                         true -> Acc;
                         false ->
                             update_or_delete_index(IndexPid, Db, DI, Proc),
-                            update_local_doc(Db, Index, PurgeSeq),
                             [{Id, Rev} | Acc]
                     end
             end,
-            clouseau_rpc:set_purge_seq(IndexPid, PurgeSeq),
-            {ok, Acc0}
+            [Acc0, PurgeSeq]
```
mayya-sharipova: did you want to use a tuple here: `{Acc0, PurgeSeq}? I have not see a list to be used like this

src/dreyfus_index_updater.erl
mayya-sharipova: @chguocloudant Yes, exactly. If a doc is partially purged, if you will get both a purge_seq change (where you can always just a delete doc from clouseau), and update_seq change where you can insert a new doc to clouseau.  If during purge_index, you only delete docs, the delete/update part will not run twice

src/dreyfus_index_updater.erl
mayya-sharipova: @chguocloudant I still think this check is not necessary. First, a user wants to delete several revs of a particular doc, the user should provided them: _purge  ID:[Rev1, Rev2, Rev3]; and this will create only one record of `{PurgeSeq, _UUID, Id, _Revs}` in purge trees.
Second, even if there are several requests with the same ID, nothing will happens if you call `clouseau_rpc:delete(IndexPid, Id)` with the same `Id` several times (I think).

src/dreyfus_index_updater.erl
chguocloudant: Thank you Mayya. I will modify it.

src/dreyfus_index.erl
chguocloudant: Thank you Mayya. I will remove the log.

src/dreyfus_index_updater.erl
chguocloudant: Hi @mayya-sharipova .
1. I used to delete the index first and update(rebuild) it in PR(#11). But Robert Newson asked me to `go from its current state to the new state without an intermediate state`, and `Here we will make the intermediate state, where the Documents in the Lucene index are removed, visible to search results.` So we delete or delete index instead of delete first and then update.
2. Based 1, we found that the index maybe delete/update twice, first in `purge_index` and second in `load_docs`. And update and delete in `clouseau` is expensive so we add an `ExcludeList` to avoid the unnecessary operations.

src/dreyfus_index_updater.erl
chguocloudant: Hi @mayya-sharipova .
The `Acc` is a list of {Id, Rev}. And the `Rev` is not the purged `Rev`, it is the final `Rev`.
For example. A doc(DocA) has Rev1, Rev2, Rev3, and purged Rev1 and Rev2. Then {DocA, Rev3} will be kept in `Acc`.
And if we call `clouseau_rpc:delete` twice with same Id, I am not sure what will happen in `Lucene` but `IndexWriter.deleteDocuments` is called twice directly.

src/dreyfus_index.erl
chguocloudant: Hi Mayya. 
clouseau will return the default value 0 if it has not yet store a purge_seq.
And if the local purge doc is not exist here, it means purge_index is not called. The purge_seq should be 0 here.
And I am not sure the means "get purge_seq from a database". The purge seq is not kept in database now.
And the line #124 should be move to line #126, thank you, I will move this line.

src/dreyfus_index_updater.erl
mayya-sharipova: @chguocloudant Thanks for explanation, your current code makes more sense now.  Please keep it.

src/dreyfus_fabric_cleanup.erl
```diff
@@ -18,13 +18,15 @@
 -include("dreyfus.hrl").
 -include_lib("mem3/include/mem3.hrl").
 -include_lib("couch/include/couch_db.hrl").
+-include_lib("couch_mrview/include/couch_mrview.hrl").
```
davisp: What's this import for? This links dreyfus to couch_mrview which is not a good idea.

src/dreyfus_index.erl
```diff
@@ -369,3 +383,23 @@ group2_int(Pid, QueryArgs0) ->
 
 info_int(Pid) ->
     clouseau_rpc:info(Pid).
+
+verify_index_exists(Options) ->
+    DbName = dreyfus_util:get_value_from_options(<<"db_name">>, Options),
```
davisp: These keys are different than what the change to couch_mrview added. We should settle on the naming scheme (even though this would work fine, consistency is always better).

src/dreyfus_util.erl
```diff
@@ -166,6 +167,46 @@ time(Metric, {M, F, A}) when is_list(Metric) ->
         couch_stats:update_histogram([dreyfus | Metric],  Length)
     end.
 
+utc_timestamp() ->
```
davisp: We should consider adding this function to couch_util since its used multiple times. We may also want to check if something like this already exists.

src/dreyfus_util.erl
```diff
@@ -166,6 +167,46 @@ time(Metric, {M, F, A}) when is_list(Metric) ->
...
+get_value_from_options(Key, Options) ->
+    case couch_util:get_value(Key, Options) of
+        undefined ->
+            Reason = binary_to_list(Key) ++ " must exist in Options.",
+            throw({bad_request, Reason});
+        Value -> Value
+    end.
+
+get_local_purge_doc_id(Sig) ->
+    <<?LOCAL_DOC_PREFIX, "purge-search-", Sig/binary>>.
```
davisp: Should we call this dreyfus to be a bit more specific than search?

src/dreyfus_index_updater.erl
```diff
@@ -77,15 +129,14 @@ load_docs(FDI, _, {I, IndexPid, Db, Proc, Total, LastCommitTime}=Acc) ->
             [Fields|_] = proc_prompt(Proc, [<<"index_doc">>, Json]),
             Fields1 = [list_to_tuple(Field) || Field <- Fields],
             case Fields1 of
-                [] -> ok = clouseau_rpc:delete(IndexPid, Id);
-                _  -> ok = clouseau_rpc:update(IndexPid, Id, Fields1)
+                [] ->
```
davisp: Undo this hunk since its not changing anything except formatting.

src/dreyfus_index_updater.erl
```diff
@@ -64,10 +68,58 @@ update(IndexPid, Index) ->
...
+            {ok, setelement(1, Acc, I+1)}
+    end.
+
+purge_index(Db, IndexPid, Index) ->
+    {ok, IdxPurgeSeq} = clouseau_rpc:get_purge_seq(IndexPid),
+    Proc = get_os_process(Index#index.def_lang),
+    try
+        true = proc_prompt(Proc, [<<"add_fun">>, Index#index.def]),
+        FoldFun = fun({PurgeSeq, _UUID, Id, _Revs}, {Acc, _}) ->
+            Acc0 = case couch_db:get_full_doc_info(Db, Id) of
```
davisp, line 95: You shouldn't need to be loading the document info here at all. You just need to delete the index entry if it exists and not cause an error if it doesn't exist. That way your update code stays the same and your change will be considerably smaller and less prone to breakage. Adding data to the index while operating over the purge sequences isn't how this was designed so there may very well be consequences to adding data based on two orderings that haven't been thought of.

davisp (changes requested):
This is mostly good. I think the operations during purge are wrong though. We've never added entries to an index during purge and that's not how its designed so doing so worries me quite a bit. I can't think of how it would fail but I can't convince myself there's not a bug there. However if we go the normal route of just removing any entries there then we don't have to worry about it.

src/dreyfus_fabric_cleanup.erl
chguocloudant: Thanks, it should be removed. I will modify it.

src/dreyfus_index_updater.erl
chguocloudant: Thank you. I will modify it.

src/dreyfus_util.erl
chguocloudant: Thank you. I will modify "search" to "dreyfus".

src/dreyfus_index.erl
chguocloudant: I will keep them same to hastings and couch_mrview.

src/dreyfus_index.erl
chguocloudant: Understand now and committed it already.

src/dreyfus_util.erl
chguocloudant: We didn't find any timestamp function in couch_util yet. I will modify it after it added in couch_util.

src/dreyfus_index_updater.erl
chguocloudant: Hi @davisp . 
The last PR (#11) is reviewed by @rnewson . And we realized that  if we only delete index in purging, we will make a intermediate state where the Documents in the Lucene index are removed, visible to search results.

rnewson: @davisp This is to prevent the index passing through an intermediate stage where the old rows are purged but the replacement rows are not yet inserted, a state that it should not be in, even temporarily. This wouldn't matter if clustered purge was just for the very rare genuine cases of purge, but we intend to use it a lot, and as part of 'hard delete'.

PR #14, opened 2017-03-20
Couchdb 3326 clustered purge

chguocloudant:
## Overview

The primary purpose of clustered purge is to clean databases that have multiple deleted tombstones or single documents that contain large numbers of conflicts. But it can also be used to purge any document (deleted or non-deleted) with any number of revisions.

With PR https://github.com/apache/couchdb/pull/1187, the purge API and update on secondary index are implemented. The current PR is in charge of updating search index according to purge tree and record purge history in local purge document to make sure that the search index update is not out of date.

## Testing recommendations

https://github.com/cloudant-labs/dreyfus/blob/f72d9c04b5262fe471b539620bc688785f0d9fab/test/dreyfus_purge_test.erl

## JIRA issue number
https://issues.apache.org/jira/browse/COUCHDB-3326

## Related Issues or Pull Requests

https://github.com/apache/couchdb/pull/1186
https://github.com/apache/couchdb/pull/1187

## Checklist

- [X] Code is written and works correctly;
- [X] Changes are covered by tests;
- [X] Documentation reflects the changes;

### DIFF
--- a/src/dreyfus_fabric_cleanup.erl
+++ b/src/dreyfus_fabric_cleanup.erl
@@ -25,6 +25,7 @@ go(DbName) ->
     {ok, DesignDocs} = fabric:design_docs(DbName),
     ActiveSigs = lists:usort(lists:flatmap(fun active_sigs/1,
         [couch_doc:from_json_obj(DD) || DD <- DesignDocs])),
+    cleanup_local_purge_doc(DbName, ActiveSigs),
     clouseau_rpc:cleanup(DbName, ActiveSigs),
     ok.
 
@@ -35,3 +36,39 @@ active_sigs(#doc{body={Fields}}=Doc) ->
          {ok, Index} = dreyfus_index:design_doc_to_index(Doc, IndexName),
          Index#index.sig
      end || IndexName <- IndexNames].
+
+cleanup_local_purge_doc(DbName, ActiveSigs) ->
+    {ok, BaseDir} = clouseau_rpc:get_root_dir(),
+    DbNamePattern = <<DbName/binary, ".*">>,
+    Pattern0 = filename:join([BaseDir, "shards", "*", DbNamePattern, "*"]),
+    Pattern = binary_to_list(iolist_to_binary(Pattern0)),
+    DirListStrs = filelib:wildcard(Pattern),
+    DirList = [iolist_to_binary(DL) || DL <- DirListStrs],
+    LocalShards = mem3:local_shards(DbName),
+    ActiveDirs = lists:foldl(fun(LS, AccOuter) ->
+        lists:foldl(fun(Sig, AccInner) ->
+            DirName = filename:join([BaseDir, LS#shard.name, Sig]),
+            [DirName | AccInner]
+        end, AccOuter, ActiveSigs)
+    end, [], LocalShards),
+
+    DeadDirs = DirList -- ActiveDirs,
+    lists:foldl(fun(IdxDir) ->
+        Sig = dreyfus_util:get_signature_from_idxdir(IdxDir),
+        case Sig of undefined -> ok; _ ->
+            DocId = dreyfus_util:get_local_purge_doc_id(Sig),
+            LocalShards = mem3:local_shards(DbName),
+            lists:foldl(fun(LS, _AccOuter) ->
+                ShardDbName = LS#shard.name,
+                {ok, ShardDb} = couch_db:open_int(ShardDbName, []),
+                case couch_db:open_doc(ShardDb, DocId, []) of
+                    {ok, LocalPurgeDoc} ->
+                        couch_db:update_doc(ShardDb,
+                            LocalPurgeDoc#doc{deleted=true}, [?ADMIN_CTX]);
+                    {not_found, _} ->
+                        ok
+                end,
+                couch_db:close(ShardDb)
+            end, [], LocalShards)
+        end
+    end, [], DeadDirs).
--- a/src/dreyfus_index.erl
+++ b/src/dreyfus_index.erl
@@ -109,6 +109,7 @@ init({DbName, Index}) ->
             case couch_db:open_int(DbName, []) of
                 {ok, Db} ->
                     try couch_db:monitor(Db) after couch_db:close(Db) end,
+                    dreyfus_util:maybe_create_local_purge_doc(Db, Pid, Index),
                     proc_lib:init_ack({ok, self()}),
                     gen_server:enter_loop(?MODULE, [], State);
                 Error ->
--- a/src/dreyfus_index_updater.erl
+++ b/src/dreyfus_index_updater.erl
@@ -31,8 +31,9 @@ update(IndexPid, Index) ->
     erlang:put(io_priority, {view_update, DbName, IndexName}),
     {ok, Db} = couch_db:open_int(DbName, []),
     try
-        %% compute on all docs modified since we last computed.
-        TotalChanges = couch_db:count_changes_since(Db, CurSeq),
+        TotalUpdateChanges = couch_db:count_changes_since(Db, CurSeq),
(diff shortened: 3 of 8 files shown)


jiangphcn:
The following changes are also reflected in current PR.
- https://github.com/cloudant-labs/dreyfus/pull/11
- https://github.com/cloudant-labs/dreyfus/pull/12

src/dreyfus_fabric_cleanup.erl
```diff
@@ -35,3 +36,33 @@ active_sigs(#doc{body={Fields}}=Doc) ->
...
+        lists:foldl(fun(Sig, AccInner) ->
+            DirName = filename:join([BaseDir, LS#shard.name, Sig]),
+            [DirName | AccInner]
+        end, AccOuter, ActiveSigs)
+    end, [], LocalShards),
+
+    DeadDirs = DirList -- ActiveDirs,
+    DeleteDocs = lists:map(fun(IdxDir) ->
+        IdxDirList = filename:split(IdxDir),
+        [Sig] = lists:nthtail(length(IdxDirList)-1, IdxDirList),
```
davisp: lists:last would be better here.

src/dreyfus_index_updater.erl
```diff
@@ -64,10 +70,64 @@ update(IndexPid, Index) ->
...
+    %% Force a commit every minute
+    case timer:now_diff(Now = now(), LastCommitTime) >= 60000000 of
+        true ->
+            ok = clouseau_rpc:commit(IndexPid, Seq),
+            {ok, {I+1, IndexPid, Db, Proc, Total, Now}};
+        false ->
+            {ok, setelement(1, Acc, I+1)}
+    end.
+
+purge_index(Db, IndexPid, Index) ->
```
davisp, line 91: This function is all wrong. There's no reason to run docs through the index function during a purge. This logic should just use fold_purge_infos and then `clouseau_rpc:delete(IndexPid, Id)` for each row. If a document was partially purged it'll be re-inserted when we process the change normally because partially purged documents are always re-inserted into the changes feed.

src/dreyfus_util.erl
```diff
@@ -166,6 +174,88 @@ time(Metric, {M, F, A}) when is_list(Metric) ->
...
+            clouseau_rpc:set_purge_seq(IndexPid, DbPurgeSeq),
+            DocContent = dreyfus_util:get_local_purge_doc_body(
+                Db, DocId, DbPurgeSeq, Index),
+            couch_db:update_doc(Db, DocContent, []);
+        _ ->
+            ok
+    end.
+
+get_local_purge_doc_id(Sig) ->
+    Version = "v" ++ config:get("purge", "version", "1") ++ "-",
```
davisp: Remove this version stuff. Its not super useful. If a purge client wants to do versioning it should be in the doc body. Nothing handles this logic so its just adding misleading complexity.

src/dreyfus_util.erl
```diff
@@ -166,6 +174,88 @@ time(Metric, {M, F, A}) when is_list(Metric) ->
...
+    ?l2b(?LOCAL_DOC_PREFIX ++ "purge-" ++ Version ++ "dreyfus-" ++ Sig).
+
+get_local_purge_doc_body(Db, LocalDocId, PurgeSeq, Index) ->
```
davisp, line 280: Seems like we should probably move the verify_index_exists function to this module so that we don't have to look back and forth between the two to make sure names are the same in both places.

src/dreyfus_index.erl
```diff
@@ -118,6 +119,7 @@ init({DbName, Index}) ->
             proc_lib:init_ack(Error)
     end.
 
+
```
davisp: Drop this whitespace only hunk

src/dreyfus_fabric_cleanup.erl
```diff
@@ -35,3 +36,33 @@ active_sigs(#doc{body={Fields}}=Doc) ->
...
+        case re:run(Sig, "^[a-fA-F0-9]+$" ,[{capture, none}]) of
```
davisp: Should we do like for views and move the "sig_from_filename" to dreyfus_util?

davisp (changes requested):
Other than the updater logic this is mostly just some style change requests.

test/dreyfus_purge_test.erl
```diff
@@ -0,0 +1,794 @@
...
+% the License.
+
+-module(dreyfus_purge_test).
+
+-include_lib("couch/include/couch_db.hrl").
+-include_lib("dreyfus/include/dreyfus.hrl").
+-include_lib("couch/include/couch_eunit.hrl").
+-include_lib("mem3/include/mem3.hrl").
+
+-export([test_purge_single/0, test_purge_multiple/0, test_purge_multiple2/0,
```
davisp, line 21: Forgot to expand this diff on my first review, sorry.

Tests are auto-exported if we're using eunit correct. The test suite for hastings does this properly and we should update these to match that approach.

test/dreyfus_purge_test.erl
```diff
@@ -0,0 +1,794 @@
...
+    DbName = db_name(),
+    create_db_docs(DbName),
+
+    %first search request
+    {ok, _, HitCount1, _, _, _} = dreyfus_search(DbName, Query),
+
+    ?assertEqual(HitCount1, 5),
+
+    %purge 5 docs
+    purge_docs(DbName, [<<"apple">>, <<"tomato">>, <<"cherry">>, <<"haw">>, <<"strawberry">>]),
```
davisp: There's a lot of super long lines in these tests that should all be fixed to be under 80 columns wide.

test/dreyfus_purge_test.erl
```diff
@@ -0,0 +1,794 @@
...
+%private API
+db_name() ->
+    Nums = tuple_to_list(erlang:now()),
+    Prefix = "test-db",
+    Suffix = lists:concat([integer_to_list(Num) || Num <- Nums]),
+    list_to_binary(Prefix ++ "-" ++ Suffix).
+
+purge_docs(DBName, DocIds) ->
+    IdsRevs = [{DocId, [get_rev(DBName, DocId)]} || DocId <- DocIds],
+    {ok, _} = fabric:purge_docs(DBName, IdsRevs, []),
```
davisp, line 713: There's no need to be going through fabric for these purge tests.

test/dreyfus_purge_test.erl
```diff
@@ -0,0 +1,794 @@
...
+
+wait_for_replicate(_, _, _, 0) ->
+    couch_log:notice("[~p] wait time out", [?MODULE]),
+    ok;
+wait_for_replicate(DbName, DocIds, ExpectRevCount ,TimeOut) when is_list(DocIds) ->
+    [wait_for_replicate(DbName, DocId, ExpectRevCount ,TimeOut) || DocId <- DocIds];
+wait_for_replicate(DbName, DocId, ExpectRevCount ,TimeOut) ->
+    FDI = fabric:get_full_doc_info(DbName, DocId, []),
+    #doc_info{ revs = Revs } = couch_doc:to_doc_info(FDI),
+    if
```
davisp: An odd use of `if` that's not normal. I'd make this a case statement since that's our usual approach for these types of cases.

davisp (changes requested):
Managed to miss the unit test diff so added some more comments there.

src/dreyfus_index_updater.erl
jiangphcn: For hastings, the index will be directly deleted/updated in easton while there is cycle about 30 seconds in clouseau side to perform final commit. So we are trying to perform batch processing here to purge indexes. More previous discussion can be found at https://github.com/cloudant-labs/dreyfus/pull/14/files#diff-8b1222f75a3b6a12a1a0ca49c6c5458aR91

davisp (approved):
+1

PR #15, opened 2017-04-11
Fix shard replacement

mayya-sharipova:
Counters were not stored as proper orddict (in a sorted manner).
Because of this when some Workers replied, they could not be found in
Counters and updated. dreyfus_fabric_search will keep waiting for
these Counters, which would lead to timeout on a request.

BugzID: 84146


davisp:
+1

rnewson:
+1

PR #16, opened 2017-05-09
Stop search indexer when it is deleted.

brkolla:
Allow dreyfus to listen to ddoc_updated events and check if the index
is deleted as a result of update to it's design document. If the index
is deleted then it will stop the index if waitlist is empty, if the
waitlist is not empty then it will wait for a minute before shutting
down.

BugzId:85718


brkolla:
@rnewson Can you please review the dreyfus code change
@iilyak Can you please review the unit tests

src/dreyfus_index_manager.erl
```diff
@@ -113,25 +115,48 @@ handle_db_event(DbName, created, _St) ->
...
     {ok, nil};
+handle_db_event(DbName, {ddoc_updated, DDocId}, _St) ->
+    DDocResult = couch_util:with_db(DbName, fun(Db) ->
+        couch_db:open_doc(Db, DDocId, [ejson_body, ?ADMIN_CTX])
+    end),
+    couch_log:info("Received ddoc_updated event for ~s",[DDocId]),
+    DbShards = [mem3:name(Sh) || Sh <- mem3:local_shards(mem3:dbname(DbName))],
+    lists:foreach(fun(DbShard) ->
+        lists:foreach(fun({_DbShard, {_DDocId, Sig}}) ->
+        case ets:lookup(?BY_SIG, {DbShard, Sig}) of
```
iilyak: Indent `case`

src/dreyfus_index_manager.erl
```diff
@@ -113,25 +115,48 @@ handle_db_event(DbName, created, _St) ->
...
+            [{_, IndexPid}] ->
+                (catch gen_server:cast(IndexPid, {ddoc_updated, DDocResult}));
+            [] ->
+                ok
+        end
+    end, ets:match_object(?BY_DB, {DbShard, {DDocId, '$1'}})) end, DbShards),
```
iilyak: could you move second `end` to another line?

src/dreyfus_index_manager.erl
```diff
@@ -113,25 +115,48 @@ handle_db_event(DbName, created, _St) ->
 handle_db_event(DbName, deleted, _St) ->
     gen_server:cast(?MODULE, {cleanup, DbName}),
     {ok, nil};
+handle_db_event(DbName, {ddoc_updated, DDocId}, _St) ->
```
iilyak: You are subscribing for `all_dbs` [here](https://github.com/cloudant-labs/dreyfus/blob/0739e2329877ea9222c8b00340c9d3b37b423220/src/dreyfus_index_manager.erl#L48). This means that you would get non-clustered databases as well. You should distinguish these two cases. Which can be done by matching on `<<"shards/", _/binary>> = DbName`.

test/search_index_ddoc_update_tests.erl
```diff
@@ -0,0 +1,277 @@
...
+-define(PASS, "pass").
+-define(AUTH, {basic_auth, {?USER, ?PASS}}).
+
+
+start() ->
+    Ctx = test_util:start_couch([chttpd, dreyfus]),
+    ok = config:set("admins", ?USER, ?PASS, _Persist=false),
+    Ctx.
+
+setup(PortType) ->
```
iilyak: dreyfus is not accessible on backdoor interface. Therefore there is no need to pass PortType and use `foreachx` in the test.

test/search_index_ddoc_update_tests.erl
```diff
@@ -0,0 +1,277 @@
...
+        _ ->
+            ok
+    end.
+
+get_index_by_sig_from_ets(DDocId, Sig) ->
+    Indexes = ets:match_object(
+        dreyfus_by_db, {'$1', {DDocId, Sig}}),
+    lists:foldl(fun({DbName, {_DDocId, _Sig}}, Acc) ->
+        case ets:lookup(dreyfus_by_sig, {DbName, Sig}) of
+            [{_, Pid}] -> [Pid| Acc];
```
iilyak: There is no guarantee that `dreyfus_by_sig` ets table contains Pid of an indexer process. There is a case when it would contain a Pid of an opener process [see here](https://github.com/cloudant-labs/dreyfus/blob/0739e2329877ea9222c8b00340c9d3b37b423220/src/dreyfus_index_manager.erl#L60).
In order to resolve this ambiguity you could use the approach [similar to the this](https://github.com/cloudant/couchdb/blob/6033c33bfaf8bb22c762eb01aedde57b6973fbd2/src/couch_index/test/couch_index_ddoc_updated_tests.erl#L127).

test/search_index_ddoc_update_tests.erl
```diff
@@ -0,0 +1,277 @@
...
+            {<<"_rev">>, ?b2l(couch_doc:rev_to_str(DDoc#doc.revs))},
+            {<<"indexes">>, {[
+                {<<"default">>, {[
+                    {<<"index">>, <<"function(doc){if(doc.blah){index(\\\"default\\\", doc.blah);}}">>}
+                ]}}
+            ]}}
+        ]},
+    UpdatedDDocJsonObj.
+
+ensureIndexIsOpened(DbName, DDocId, IndexName) ->
```
iilyak: we don't use camel case naming convention in erlang

test/search_index_ddoc_update_tests.erl
```diff
@@ -0,0 +1,277 @@
...
+     ?_test(begin
+         ok = meck:reset(clouseau_rpc),
+         ok = meck:reset(dreyfus_index),
+         ok = create_doc(PortType, ?l2b(DbName), <<"doc_id">>, {[]}),
+         ReqUrl = Host ++ "/" ++ DbName ++ "/_design/searchddoc/_search/other?q=*:*",
+         {ok, Status, _Headers, _Body} =
+              test_request:get(ReqUrl, [?AUTH]),
+          ?assertEqual(200, Status),
+          DDocId = "_design/searchddoc",
+          Sig = ensureIndexIsOpened(DbName, DDocId, "other"),
```
iilyak: You want to make sure that the following conditions are satisfied:
 1. the deleted index is not in ets table
 2. the old index Pid is not running after deletion of search index.

Also keep in mind that you would have multiple indexers running (one per database shard).
I would suggest to use Mayya's approach [from here](https://github.com/cloudant/couchdb/blob/6033c33bfaf8bb22c762eb01aedde57b6973fbd2/src/couch_index/test/couch_index_ddoc_updated_tests.erl#L71-L102)

test/search_index_ddoc_update_tests.erl
```diff
@@ -0,0 +1,277 @@
...
+              }}).
+
+make_test_case(Mod, Funs) ->
+    {
+        lists:flatten(io_lib:format("~s", [Mod])),
+        {foreachx, fun setup/1, fun teardown/2, [{Mod, Fun} || Fun <- Funs]}
+    }.
+
+should_stop_indexing_after_deleting_index(PortType, {Host, DbName}) ->
+    {timeout, 60,
```
iilyak: Why you had to increase timeout?

iilyak (changes requested):
I added some comments

brkolla:
Thanks @iilyak. Will check them out and make the suggested changes as needed.

test/search_index_ddoc_update_tests.erl
brkolla: In this case, I am issuing a search request, before I call this method.. so it would have opened the index pid before it executes this code..

test/search_index_ddoc_update_tests.erl
brkolla: fixed

test/search_index_ddoc_update_tests.erl
brkolla: Yes, the code in [here](https://github.com/cloudant-labs/dreyfus/blob/0739e2329877ea9222c8b00340c9d3b37b423220/src/dreyfus_index_manager.erl#L123-L132) is taking care of handling the event in all db shards.

test/search_index_ddoc_update_tests.erl
brkolla: I believe the test was failing if I didn't increase timeout.. May be because I am waiting for the index to be closed before I verify the index is indeed closed.. And that wait seems like is causing the problem.

test/search_index_ddoc_update_tests.erl
brkolla: Also, if the index pid is not there in ets means it is not running right? Isn't that enough to verify that the index pid for the old index is not running? What's the scenario where the index pid would not be in ets but would still be running.. I don't think that's possible..

src/dreyfus_index.erl
```diff
@@ -40,6 +40,8 @@
     waiting_list=[]
 }).
 
+-define(INDEX_SHUTDOWN_DELAY, 60000). % 1 minute
```
rnewson: needs to be configurable.

src/dreyfus_index.erl
```diff
@@ -157,9 +159,43 @@ handle_call(info, _From, State) -> % obsolete
...
+        {ok, DDoc} ->
+            Index = get_index_with_sig(DDoc, CurrentSig),
+            case Index of
+                [] ->
+                    true;
+                _ ->
+                    false
+            end
+        end,
+        case IndexDeleted of
```
rnewson: indentation is misleadingly wrong here.

src/dreyfus_index.erl
```diff
@@ -157,9 +159,43 @@ handle_call(info, _From, State) -> % obsolete
     Reply = info_int(State#state.index_pid),
     {reply, Reply, State}.
 
+handle_cast({ddoc_updated, DDocResult}, State) ->
+    #state{dbname = DbName, index=#index{sig=CurrentSig}, waiting_list = WaitList} = State,
+    IndexDeleted = case DDocResult of
```
rnewson: extract this into a function for clarity.

src/dreyfus_index.erl
```diff
@@ -157,9 +159,43 @@ handle_call(info, _From, State) -> % obsolete
...
+        {not_found, deleted} ->
+            true;
+        {ok, DDoc} ->
+            Index = get_index_with_sig(DDoc, CurrentSig),
+            case Index of
+                [] ->
+                    true;
+                _ ->
```
rnewson: match something explicitly here, a non-empty list, for example. we don't expect any other result from get_index_with_sig, so don't have clauses for them.

src/dreyfus_index.erl
```diff
@@ -157,9 +159,43 @@ handle_call(info, _From, State) -> % obsolete
...
+                    [] ->
+                        self() ! {index_deleted},
+                        {noreply, State};
+                    _ ->
+                        erlang:send_after(?INDEX_SHUTDOWN_DELAY, self(), {index_deleted}),
+                        {noreply, State}
+                end;
+            false ->
+                {noreply, State}
+        end;
```
rnewson: since we always return {noreply, State}, move it to after the case statement so we say it once.

src/dreyfus_index.erl
```diff
@@ -157,9 +159,43 @@ handle_call(info, _From, State) -> % obsolete
...
+                    false
+            end
+        end,
+        case IndexDeleted of
+            true ->
+                case WaitList of
+                    [] ->
+                        self() ! {index_deleted},
+                        {noreply, State};
+                    _ ->
```
rnewson: again, the only other thing we expect here is a non-empty list, say so.

src/dreyfus_index.erl
```diff
@@ -157,9 +159,43 @@ handle_call(info, _From, State) -> % obsolete
...
+                        erlang:send_after(?INDEX_SHUTDOWN_DELAY, self(), {index_deleted}),
```
rnewson: why?

src/dreyfus_index.erl
```diff
@@ -157,9 +159,43 @@ handle_call(info, _From, State) -> % obsolete
...
+                        {noreply, State}
+                end;
+            false ->
+                {noreply, State}
+        end;
+
 handle_cast(_Msg, State) ->
     {noreply, State}.
 
+handle_info({index_deleted}, State) ->
```
rnewson: this could just be `index_deleted`.

src/dreyfus_index.erl
```diff
@@ -222,6 +258,22 @@ code_change(_OldVsn, State, _Extra) ->
...
+            {IndexNames, _} = lists:unzip(IndexList),
+            lists:flatmap(
+                fun(IndexName) ->
+                    case (catch design_doc_to_index(Doc, IndexName)) of
+                        {ok, #index{sig=Sig_New}=Index} when Sig_New =:= Sig -> [Index];
+                        _ -> []
+                    end
+                end,
+                IndexNames);
+        _ -> []
```
rnewson: what else could it be? we should avoid defensive programming, it causes more problems than it solves.

src/dreyfus_index.erl
```diff
@@ -222,6 +258,22 @@ code_change(_OldVsn, State, _Extra) ->
...
+get_index_with_sig(#doc{body={Fields}}=Doc, Sig) ->
+    RawIndexes = couch_util:get_value(<<"indexes">>, Fields, {[]}),
+    case RawIndexes of
+        {IndexList} when is_list(IndexList) ->
+            {IndexNames, _} = lists:unzip(IndexList),
+            lists:flatmap(
+                fun(IndexName) ->
+                    case (catch design_doc_to_index(Doc, IndexName)) of
+                        {ok, #index{sig=Sig_New}=Index} when Sig_New =:= Sig -> [Index];
+                        _ -> []
```
rnewson: list the other cases.

src/dreyfus_index.erl
```diff
@@ -222,6 +258,22 @@ code_change(_OldVsn, State, _Extra) ->
...
 
+get_index_with_sig(#doc{body={Fields}}=Doc, Sig) ->
+    RawIndexes = couch_util:get_value(<<"indexes">>, Fields, {[]}),
+    case RawIndexes of
+        {IndexList} when is_list(IndexList) ->
+            {IndexNames, _} = lists:unzip(IndexList),
+            lists:flatmap(
+                fun(IndexName) ->
+                    case (catch design_doc_to_index(Doc, IndexName)) of
+                        {ok, #index{sig=Sig_New}=Index} when Sig_New =:= Sig -> [Index];
```
rnewson: I think you could just do `{ok, #index{sig=Sig}=Index}` here.

src/dreyfus_index_manager.erl
```diff
@@ -113,25 +115,49 @@ handle_db_event(DbName, created, _St) ->
...
+    DDocResult = couch_util:with_db(DbName, fun(Db) ->
+        couch_db:open_doc(Db, DDocId, [ejson_body, ?ADMIN_CTX])
+    end),
+    couch_log:info("Received ddoc_updated event for ~s",[DDocId]),
+    DbShards = [mem3:name(Sh) || Sh <- mem3:local_shards(mem3:dbname(DbName))],
+    lists:foreach(fun(DbShard) ->
+        lists:foreach(fun({_DbShard, {_DDocId, Sig}}) ->
+            case ets:lookup(?BY_SIG, {DbShard, Sig}) of
+                [{_, IndexPid}] ->
+                    (catch gen_server:cast(IndexPid, {ddoc_updated, DDocResult}));
```
rnewson: why catch?

src/dreyfus_index.erl
```diff
@@ -222,6 +258,22 @@ code_change(_OldVsn, State, _Extra) ->
...
 % private functions.
 
+get_index_with_sig(#doc{body={Fields}}=Doc, Sig) ->
+    RawIndexes = couch_util:get_value(<<"indexes">>, Fields, {[]}),
+    case RawIndexes of
+        {IndexList} when is_list(IndexList) ->
+            {IndexNames, _} = lists:unzip(IndexList),
+            lists:flatmap(
+                fun(IndexName) ->
+                    case (catch design_doc_to_index(Doc, IndexName)) of
```
rnewson: why catch?

src/dreyfus_index_manager.erl
```diff
@@ -113,25 +115,49 @@ handle_db_event(DbName, created, _St) ->
...
+                [] ->
+                    ok
+            end
+        end, ets:match_object(?BY_DB, {DbShard, {DDocId, '$1'}}))
```
rnewson: what's the $1 for if we don't apply a match condition?

src/dreyfus_index_manager.erl
```diff
@@ -113,25 +115,49 @@ handle_db_event(DbName, created, _St) ->
...
 
-add_to_ets(Pid, DbName, Sig) ->
+add_to_ets(Pid, DbName, DDocId, Sig) ->
     true = ets:insert(?BY_PID, {Pid, {DbName, Sig}}),
-    true = ets:insert(?BY_SIG, {{DbName, Sig}, Pid}).
+    true = ets:insert(?BY_SIG, {{DbName, Sig}, Pid}),
+    true = ets:insert(?BY_DB, {DbName, {DDocId, Sig}}).
 
 delete_from_ets(Pid, DbName, Sig) ->
+    case ets:match_object(?BY_DB, {DbName, {'$1', Sig}}) of
```
rnewson: what's the $1 for if we don't apply a match condition?

src/dreyfus_index.erl
brkolla: Basically if the waitlist is empty, it could either be waiting for the indexing to complete or it could be waiting for the end user request. Other option is to check and find out if the waitlist is the result of the end user request or the Ken request, but I don't think it's worth the effort.. just wait for one minute to let the index to finish the end user requests if any in the wait list.

src/dreyfus_index.erl
brkolla: I am trying to compare the if the index list has the index matching the Sig that was passed in as a parameter to this function.

src/dreyfus_index_manager.erl
brkolla: To be honest, I just copied the code from couch_index_server and hence the catch.. I thought the motivation was to not kill the index manager if there was an error sending message to search index.. But only time this could raise an error is when something really bad happens to the node.. I will take out the exception..

src/dreyfus_index_manager.erl
brkolla: I followed the same concept as in [couch_index_server](https://github.com/cloudant/couchdb-couch-index/blob/master/src/couch_index_server.erl#L262). I thought this would basically find all the search indices opened for the given ddoc matching DDocId. Am I missing something here?

src/dreyfus_index_manager.erl
brkolla: This is to find the design doc id for a given DB and Search index, so that I can remove it from the dreyfus_by_db ets.

src/dreyfus_index.erl
brkolla: looks like we don't have any config for dreyfus, if I make this configurable then do we need to add a new config file?
Not sure if it's worth the effort to make this configurable.

test/search_index_ddoc_update_tests.erl
brkolla: I tried to update this unit test without the PortType and couldn't make it work.. Can you please show me an example unit test code that I can follow? I want to use this in the mean time as it's doing the intended job.

brkolla:
@rnewson @iilyak Have updated the code to reflect the changes as per your suggestions and also replied to your comments. Can you please take another attempt to review?

src/dreyfus_index.erl
rnewson: I mean `config:get`, that's available to all applications.

src/dreyfus_index.erl
```diff
@@ -40,6 +40,9 @@
     waiting_list=[]
 }).
 
+-define(INDEX_SHUTDOWN_DELAY,
+    config:get("dreyfus", "index_shutdown_delay", 60000)). % 1 minute
```
rnewson: config:get returns a string, so the default needs to be a string, or use get_integer. And move this to a function please.

brkolla:
@rnewson, Done. Moved the config:get to a function and also changed it to use get_integer.

test/search_index_ddoc_update_tests.erl
```diff
@@ -0,0 +1,277 @@
...
+
+ensure_index_is_closed(DbName, DDocId, IndexName, Sig) ->
+    NotAnIndex = getIndex(DbName, ?l2b(DDocId), ?l2b(IndexName)),
+    ?assertEqual(ok, NotAnIndex),
+    (catch wait_for_indices_to_stop(?l2b(DbName))),
+    IndexPidsFromEts = get_index_by_sig_from_ets(?l2b(DDocId), Sig),
+    ?assertEqual(0, length(IndexPidsFromEts)),
+    ok.
+
+getIndex(DbName, DDocId, IndexName) ->
```
iilyak: we don't use CamelCase naming in function names. this function would be called `get_index`.

iilyak: Should we return `not_found` instead of `ok`?

test/search_index_ddoc_update_tests.erl
```diff
@@ -0,0 +1,277 @@
...
+
+ensure_index_is_opened(DbName, DDocId, IndexName) ->
+    IndexWithName = getIndex(DbName, ?l2b(DDocId), ?l2b(IndexName)),
+    ?assertMatch(#index{sig=_Sig}, IndexWithName),
+    [#index{sig=Sig}] = [IndexWithName],
+    IndexPidsFromEts = get_index_by_sig_from_ets(?l2b(DDocId), Sig),
+    ?assert(length(IndexPidsFromEts) > 0),
+    Sig.
+
+ensure_index_is_closed(DbName, DDocId, IndexName, Sig) ->
```
iilyak: How do you know that index Pids are actually killed? It seems like you rely on the info from ets. In this case if we would introduce a bug in a Pid accounting in ets tables we wouldn't be able to catch the problem.
You could pass the Pids of indexer instances as an argument to this function or make this check in the test case itself.

test/search_index_ddoc_update_tests.erl
```diff
@@ -0,0 +1,277 @@
...
+          Sig = ensure_index_is_opened(DbName, DDocId, "other"),
+
+          % Delete the design doc and verify that the index will be closed
+          delete_ddoc(Host, DbName, ?l2b(DDocId)),
+          ensure_index_is_closed(DbName, "_design/searchddoc", "other", Sig)
+    end)}.
+
+updated_ddoc_json_obj(DbName, DDocId) ->
+    {ok, DDoc} = fabric:open_doc(DbName, <<DDocId/binary>>, [ejson_body, ?ADMIN_CTX]),
+    % {DDocJsonObj} = ?DDOC_UPDATE,
```
iilyak: Don't forget to remove commented out code in the final version.

test/search_index_ddoc_update_tests.erl
brkolla: Not really.. The intent of that function is that it ensures that the index is closed and returning 'ok only if it succeeds that.. otherwise the assert would fail and cause the test to stop..

brkolla:
@iilyak Changed the test code to pass the indexpids and ensure that they are closed.

test/search_index_ddoc_update_tests.erl
iilyak: Here is one example of it https://github.com/apache/couchdb/blob/master/src/couch/test/couchdb_compaction_daemon_tests.erl#L62

src/dreyfus_index_manager.erl
iilyak: As Robert mentioned this shouldn't be a `'$1'`, but `'_'`. We need to bind if we want to compare that the values at different positions are the same. For example `{'$1', a, '$1'}` would match `{22, a, 22}`, but it wouldn't match `{44, a, 32}`.

src/dreyfus_index.erl
```diff
@@ -40,6 +40,8 @@
     waiting_list=[]
 }).
 
+-define(INDEX_SHUTDOWN_DELAY, index_shutdown_delay()).
```
rnewson: remove this define, just call the function where needed.

src/dreyfus_index.erl
```diff
@@ -157,9 +159,28 @@ handle_call(info, _From, State) -> % obsolete
     Reply = info_int(State#state.index_pid),
     {reply, Reply, State}.
 
+handle_cast({ddoc_updated, DDocResult}, State) ->
+    #state{index=#index{sig=CurrentSig}, waiting_list = WaitList} = State,
+    case check_if_index_is_deleted(DDocResult, CurrentSig) of
+        true ->
```
rnewson: this case statement will crash when the index isn't deleted as you haven't written the `false` case. I would also suggest collapsing this to one case statement using `{check_if_index_is_deleted(DDocResult, CurrentSig), WaitList}`

src/dreyfus_index.erl
```diff
@@ -157,9 +159,28 @@ handle_call(info, _From, State) -> % obsolete
...
+            case WaitList of
+                [] ->
+                    self() ! index_deleted;
+                [_Index] ->
```
rnewson: this case statement assumes the wait list will have zero or one items, and crashes if there is more than one waiter.

src/dreyfus_index.erl
```diff
@@ -157,9 +159,28 @@ handle_call(info, _From, State) -> % obsolete
...
+                    erlang:send_after(?INDEX_SHUTDOWN_DELAY, self(), index_deleted)
```
rnewson: it would be better to wait until the list is empty. New requests should not be possible if the index has been deleted.

src/dreyfus_index.erl
```diff
@@ -222,6 +243,35 @@ code_change(_OldVsn, State, _Extra) ->
 
 % private functions.
 
+get_index_with_sig(#doc{body={Fields}}=Doc, Sig) ->
+    RawIndexes = couch_util:get_value(<<"indexes">>, Fields, {[]}),
+    case RawIndexes of
+        {IndexList} when is_list(IndexList) ->
```
rnewson: this case statement has one clause, and crashes on anything else. If that's the intent, you don't need a case statement.

src/dreyfus_index.erl
```diff
@@ -222,6 +243,35 @@ code_change(_OldVsn, State, _Extra) ->
 
 % private functions.
 
+get_index_with_sig(#doc{body={Fields}}=Doc, Sig) ->
```
rnewson: This function only looks for an index of given signature in a single design document. A search index might be referred to by multiple design documents.

rnewson (changes requested):
There are some substantial issues to address here.

src/dreyfus_index.erl
brkolla: I thought the whole idea of this work is to stop indexing for the deleted indices which could have already triggered and in progress. Agree that there won't be any _search end point requests after the index is deleted. Basically giving 1 minute to complete the pending requests, if they are not done then it will end the request.

src/dreyfus_index.erl
brkolla: I didn't know about this case of index could be referred by multiple design documents. If that's the case I have to scan all the design docs in the db and ensure that it's not referred in other docs.

src/dreyfus_index_manager.erl
```diff
@@ -121,25 +123,49 @@ handle_db_event(DbName, created, _St) ->
...
+    DbShards = [mem3:name(Sh) || Sh <- mem3:local_shards(mem3:dbname(DbName))],
+    lists:foreach(fun(DbShard) ->
+        lists:foreach(fun({_DbShard, {_DDocId, Sig}}) ->
+            case ets:lookup(?BY_SIG, {DbShard, Sig}) of
+                [{_, IndexPid}] ->
+                    gen_server:cast(IndexPid, {ddoc_updated, DDocResult});
+                [] ->
+                    ok
+            end
+        end, ets:match_object(?BY_DB, {DbShard, {DDocId, '$1'}}))
```
iilyak: replace `'$1'` with `'_'`

test/search_index_ddoc_update_tests.erl
```diff
@@ -0,0 +1,275 @@
...
+                _ ->
+                    ok
+            end;
+        _ ->
+            ok
+    end.
+
+get_index_by_sig_from_ets(DDocId, Sig) ->
+    Indexes = ets:match_object(
+        dreyfus_by_db, {'$1', {DDocId, Sig}}),
```
iilyak: replace `'$1'` with `'_'`

src/dreyfus_index.erl
```diff
@@ -214,13 +226,65 @@ handle_info({'DOWN',_,_,Pid,Reason}, #state{
...
-terminate(_Reason, _State) ->
-    ok.
+terminate(Reason, State) ->
+    case Reason of
+        {shutdown, ddoc_updated} ->
+            send_all(State#state.waiting_list, ddoc_updated),
+            ok;
+        _ ->
+            ok
+    end.
```
mayya-sharipova: This part looks good to me

src/dreyfus_index.erl
```diff
@@ -157,9 +157,26 @@ handle_call(info, _From, State) -> % obsolete
     Reply = info_int(State#state.index_pid),
     {reply, Reply, State}.
 
+handle_cast({ddoc_updated, DDocResult},
+            #state{
+                index=#index{sig=Sig, dbname=DbName}=Index
+            }=State) ->
+    case check_if_index_is_deleted(DbName, DDocResult, Sig) of
+        true ->
+            erlang:send_after(index_shutdown_delay(), self(), index_deleted);
```
rnewson: this will crash the gen server as it returns a value that handle_cast does not expect (a TimerRef).

rnewson: what's the delay for?

rnewson: the gen server should shut down after serving last request after deletion.

src/dreyfus_index.erl
```diff
@@ -214,13 +231,65 @@ handle_info({'DOWN',_,_,Pid,Reason}, #state{
     [gen_server:reply(P, {error, Reason}) || {P, _} <- WaitList],
     {stop, normal, State}.
 
-terminate(_Reason, _State) ->
-    ok.
+terminate(Reason, State) ->
+    case Reason of
+        {shutdown, ddoc_updated} ->
+            send_all(State#state.waiting_list, ddoc_updated),
+            ok;
```
rnewson: the return value of terminate is not relevant, I would leave this ok; out.

src/dreyfus_index.erl
```diff
@@ -214,13 +231,65 @@ handle_info({'DOWN',_,_,Pid,Reason}, #state{
...
 % private functions.
+send_all(Waiters, Reply) ->
+    [gen_server:reply(From, Reply) || {From, _} <- Waiters].
+
+check_if_index_is_deleted(DbName, DDocResult, CurrentSig) ->
+    case DDocResult of
+        {not_found, deleted} ->
+            IndicesWithSig = indices_with_sig_from_other_ddocs(DbName, CurrentSig);
+        {ok, DDoc} ->
+            IndicesWithSig = indices_with_sig(DbName, DDoc, CurrentSig)
```
rnewson: this is wrong. we always need to check if _any_ ddoc refers to the index before we can consider that it's deleted.

rnewson (changes requested):

src/dreyfus_index.erl
brkolla: We are indeed doing that.. Inside the indices_with_sig, it will first check if it's deleted then it will check if it referred from other ddocs

brkolla: As per the discussion in slack, will rename/reorg this code to make it easy to read..

src/dreyfus_index.erl
```diff
@@ -214,13 +232,62 @@ handle_info({'DOWN',_,_,Pid,Reason}, #state{
...
     {ok, State}.
 
 % private functions.
+send_all(Waiters, Reply) ->
+    [gen_server:reply(From, Reply) || {From, _} <- Waiters].
+
+check_if_index_is_deleted(DbName, DDocResult, CurrentSig) ->
+    case DDocResult of
+        {not_found, deleted} ->
+            IndicesWithSig = indices_with_sig_from_other_ddocs(DbName, CurrentSig);
```
mayya-sharipova, line 252: this `indices_with_sig_from_other_ddocs` seems to be very expensive, especially doing exactly same operation for all  N shards. 
I would suggest avoid this check. First, it is expensive. Second, it seems to me that it is not very likely that the customers have exact same indexes for the same db from diff Design docs and using these Design Docs at the same time. Third, even if we close the index used by another ddoc the danger is minimal, if will be reopened with the next query/ken job.

Even if we still want to keep this check it should be initiated from a single process spawn from `dreyfus_index_manager` and then results passed to all concerning `dreyfus_index` processes.

@rnewson @brkolla  What do you think?

src/dreyfus_index.erl
```diff
@@ -157,9 +157,27 @@ handle_call(info, _From, State) -> % obsolete
     Reply = info_int(State#state.index_pid),
     {reply, Reply, State}.
 
+handle_cast({ddoc_updated, DDocResult},
+            #state{
```
rnewson: indent by 4 spaces.

rnewson: on principle, we only extract terms in the function head if we are matching on them. so pull Sig and DbName out on a separate line. This will also make the function head much smaller and it will fit on one line.

```
handle_cast({ddoc_updated, DDocResult}, #state{} = State) ->
    #index{sig = Sig, dbname = DbName} = State#state.index,
```

src/dreyfus_index.erl
```diff
@@ -157,9 +157,27 @@ handle_call(info, _From, State) -> % obsolete
...
+                index=#index{sig=Sig, dbname=DbName}
+            }=State) ->
+    case check_if_index_is_deleted(DbName, DDocResult, Sig) of
+        true ->
+            erlang:send_after(index_shutdown_delay(), self(), index_deleted),
```
rnewson: think I said it before, but an arbitrary delay doesn't seem write. If a design document is updated such that there is no reference to this index, we should either let any requests to it finish and then delete the index or just delete it. The arbitrary delay is, I'm guessing, intended to avoid the errors caused by deleting an index while it's being queried? If so, we should at least document that it's a bit hackish.

src/dreyfus_index.erl
mayya-sharipova: @brkolla  Paul D. has commented on my similar code, and it seems that I was wrong. It is common practice to have the same DDocs when updating them. So, please keep the code as before checking on `indices_with_sig_from_other_ddocs`

src/dreyfus_index.erl
mayya-sharipova: @rnewson  what is the way to "let any requests to it finish"?  How do we know that requests are finished? check if `Waiters` == [] will not work for stale requests that don't wait for indexing to catch up.

src/dreyfus_index.erl
brkolla: K, will leave it as it is then. Thx.

brkolla:
@rnewson. I have updated the changes to make changes as per your suggestions..
And in regard to the comment quoted by you (see below), I will do whatever you and @mayya-sharipova decide. I can add a comment in the code summarizing the reasoning for our approach. 
As I and @mayya-sharipova mentioned there is no easy way to identify the pids representing the end users (and waiting for indexing to catch up) vs the pids initiated from ken. We can add the comment and move on.
 
> think I said it before, but an arbitrary delay doesn't seem write. If a design document is updated such that there is no reference to this index, we should either let any requests to it finish and then delete the index or just delete it. The arbitrary delay is, I'm guessing, intended to avoid the errors caused by deleting an index while it's being queried? If so, we should at least document that it's a bit hackish.

>

PR #17, opened 2017-06-05
Add new end point to get disk size information for search index

brkolla:
Currently the only way to get the disk size information for the search
index is to use the _search_info end point. But using this point would
lead to opening the search index which is non trivial as lucene has an
overhead for opening the index.

These changes would add a new end point _search_disk_size to get the
disk size information without opening the search index. Sarnie can use
this new end point and avoid opening the search index.

BugzId:87336

### DIFF
--- a/src/dreyfus_fabric_info.erl
+++ b/src/dreyfus_fabric_info.erl
@@ -19,15 +19,15 @@
 -include_lib("mem3/include/mem3.hrl").
 -include_lib("couch/include/couch_db.hrl").
 
--export([go/3]).
+-export([go/4]).
 
-go(DbName, DDocId, IndexName) when is_binary(DDocId) ->
+go(DbName, DDocId, IndexName, InfoLevel) when is_binary(DDocId) ->
     {ok, DDoc} = fabric:open_doc(DbName, <<"_design/", DDocId/binary>>, []),
-    go(DbName, DDoc, IndexName);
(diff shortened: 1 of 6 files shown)


brkolla:
@rnewson? Can you please review these changes?

src/dreyfus_fabric_info.erl
```diff
@@ -19,15 +19,20 @@
...
-    go(DbName, DDoc, IndexName);
+    go(DbName, DDoc, IndexName, InfoLevel);
 
-go(DbName, DDoc, IndexName) ->
+go(DbName, DDoc, IndexName, InfoLevel) ->
     Shards = mem3:shards(DbName),
-    Workers = fabric_util:submit_jobs(Shards, dreyfus_rpc, info, [DDoc, IndexName]),
+    case InfoLevel of
+        info ->
+            Workers = fabric_util:submit_jobs(Shards, dreyfus_rpc, info, [DDoc, IndexName]);
```
rnewson: just pass InfoLevel here, you don't need the case statement.

rnewson (changes requested):
remove the unnecessary case statement and this is good. thanks.

src/dreyfus_fabric_info.erl
brkolla: That's what I started out with but I thought this would be more readable. Will switch it back

rnewson (approved):

PR #21, opened 2017-08-28
Stop dreyfus_index processes on ddoc update

mayya-sharipova:
## Overview
Currently when ddoc is modified, dreyfus_index and dreyfus_index_updater
processes corresponding to the previous version of ddoc will keep running
until all indexing processing initiated by them is done.
When ddoc of a big database is rapidly modified, this puts a lot
of unnecessary strain on database resources.

This commit brings the following changes:
1. When opening a dreyfus_index, always add a record
   `{DbName, {DDocId, Sig}}` to `?BY_DB`.
2. When ddoc_updated, check if there other ddocs in `?BY_DB` with the same Sig.
    If there are other, only remove `{DbName, {DDocId, Sig}}`
    	record from `?BY_DB` for this ddoc.
    If there are no, stop dreyfus_index processes:
    * all dreyfus_index processes for the prev. version of ddoc will be shutdown
    * all linked to them dreyfus_index_updater processes will die as well
    * all waiters for indexing activity to be finished will receive an immediate
	reply: ddoc_updated. Interactive query requests will get response:
	{404, <<"not_found">>, <<"Design document was updated or deleted.">>}

## Testing recommendations
`make check apps=dreyfus tests=ddoc_update_test_`

## GitHub issue number
This is to rewrite: https://github.com/cloudant-labs/dreyfus/pull/16
To follow the same style as in: https://github.com/apache/couchdb/commit/2e7ca45b364467f164d2ecbe8846878234d36a47

BugzID: 85718

### DIFF
--- a/src/dreyfus_index.erl
+++ b/src/dreyfus_index.erl
@@ -157,10 +157,9 @@ handle_call(info, _From, State) -> % obsolete
     Reply = info_int(State#state.index_pid),
     {reply, Reply, State}.
 
-
 handle_cast({ddoc_updated, DDocResult}, #state{} = State) ->
     #index{sig = Sig} = State#state.index,
-    KeepIndex = case DDocResult of
+    KeepIndexProcess = case DDocResult of
         {not_found, deleted} ->
             false;
         {ok, DDoc} ->
@@ -170,7 +169,7 @@ handle_cast({ddoc_updated, DDocResult}, #state{} = State) ->
                 (SigNew == Sig) or Acc
             end, false, Indexes)
     end,
-    case KeepIndex of
+    case KeepIndexProcess of
         false ->
             {stop, {shutdown, ddoc_updated}, State};
         true ->
@@ -178,7 +177,6 @@ handle_cast({ddoc_updated, DDocResult}, #state{} = State) ->
     end;
 handle_cast(_Msg, State) ->
     {noreply, State}.
-
 
 handle_info({'EXIT', FromPid, {updated, NewSeq}},
             #state{
@@ -234,14 +232,11 @@ handle_info({'DOWN',_,_,Pid,Reason}, #state{
     [gen_server:reply(P, {error, Reason}) || {P, _} <- WaitList],
     {stop, normal, State}.
 
-terminate(Reason, State) ->
-    case Reason of
-        {shutdown, ddoc_updated} ->
-            Waiters = State#state.waiting_list,
-            [gen_server:reply(From, ddoc_updated) || {From, _} <- Waiters];
-        _ ->
-            ok
-    end.
+terminate({shutdown, ddoc_updated}, State) ->
+    Waiters = State#state.waiting_list,
+    [gen_server:reply(From, ddoc_updated) || {From, _} <- Waiters];
+terminate(_Reason, _State) ->
+    ok.
 
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
--- a/src/dreyfus_index.erl
+++ b/src/dreyfus_index.erl
@@ -157,8 +157,28 @@ handle_call(info, _From, State) -> % obsolete
     Reply = info_int(State#state.index_pid),
     {reply, Reply, State}.
 
+
+handle_cast({ddoc_updated, DDocResult}, #state{} = State) ->
+    #index{sig = Sig} = State#state.index,
+    KeepIndex = case DDocResult of
+        {not_found, deleted} ->
+            false;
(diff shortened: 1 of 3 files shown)


src/dreyfus_index_manager.erl
```diff
@@ -121,25 +128,71 @@ handle_db_event(DbName, created, _St) ->
...
+                % single DDoc with this Sig - maybe close dreyfus_index process
+                case ets:lookup(?BY_SIG, {DbShard, Sig}) of
+                    [{_, IndexPid}] -> (catch
+                        gen_server:cast(IndexPid, {ddoc_updated, DDocResult}));
+                    [] -> []
+                end
+            end
+        end, ets:match_object(?BY_DB, {DbShard, {DDocId, '_'}}))
+    end, DbShards),
+    {ok, nil};
```
iilyak: It is confusing to return `nil`. I think it is better to return `St` as we do in the next clause.

src/dreyfus_index.erl
```diff
@@ -157,9 +157,29 @@ handle_call(info, _From, State) -> % obsolete
...
+    case KeepIndex of
+        false ->
+            {stop, {shutdown, ddoc_updated}, State};
+        true ->
+            {noreply, State}
+    end;
 handle_cast(_Msg, State) ->
     {noreply, State}.
 
+
```
tonysun83: whitespace

src/dreyfus_index.erl
```diff
@@ -214,8 +234,14 @@ handle_info({'DOWN',_,_,Pid,Reason}, #state{
     [gen_server:reply(P, {error, Reason}) || {P, _} <- WaitList],
     {stop, normal, State}.
 
-terminate(_Reason, _State) ->
```
tonysun83: maybe rewrite this as
```
terminate({shutdown, ddoc_updated}, State) ->
    Waiters = State#state.waiting_list,
    [gen_server:reply(From, ddoc_updated) || {From, _} <- Waiters];
termiante(_Reason, _State) ->
    ok.
```

src/dreyfus_index.erl
```diff
@@ -157,9 +157,29 @@ handle_call(info, _From, State) -> % obsolete
     Reply = info_int(State#state.index_pid),
     {reply, Reply, State}.
 
+
+handle_cast({ddoc_updated, DDocResult}, #state{} = State) ->
+    #index{sig = Sig} = State#state.index,
+    KeepIndex = case DDocResult of
```
tonysun83: Should this be called KeepIndexes? If I understand correctly, if one index in the new design doc has the same signature, then we don't kill the existing indexes.  So if I had 1000 indexes, and I update 999 of them with a new definition (and thus a new signature), but only 1 still was unchanged, then that would mean 999 old indexes would still keep running right?

src/dreyfus_index_manager.erl
```diff
@@ -86,6 +90,9 @@ handle_call({open_error, DbName, Sig, Error}, {OpenerPid, _}, State) ->
 
 handle_cast({cleanup, DbName}, State) ->
     clouseau_rpc:cleanup(DbName),
+    {noreply, State};
```
tonysun83, line 93: Why do we add this here?

src/dreyfus_index.erl
mayya-sharipova: @tonysun83 `KeepIndex` here means if we should keep the current `dreyfus_index` process or kill it.  It refers to a current single `dreyfus_index` process.  May be a better name could be `KeepIndexProcess`?

src/dreyfus_index_manager.erl
mayya-sharipova: @tonysun83  Are you referring to the line 93? If yes, we did not add this here, it was here before - this a standard way to finish `handle_cast`.  This line has changed as we added a new `handle_cast` clause.

src/dreyfus_index.erl
mayya-sharipova: @tonysun83 The intention is to shutdown all changed indexes, so in your case all 999 old indexes will be killed

tonysun83: @mayya-sharipova : understand now, had to re-read it again, not I got it

tonysun83:
Looks pretty good. I saw in the tests that deletion was the method to see if index processes were killed or updated. Can one we add one test where we add a design doc with two index names, then update the definition for one index, and check to see that the old process with the old signature is killed? That coverage would be great.

tonysun83:
LGTM, +1

PR #27, opened 2018-07-06
Add ability to black list indexes

tonysun83:
This new PR provides the ability add problematic indexes to a black list.

The black list is essentially a config entry that uses a list as it's value.
We can consider using hash structure later for efficiency. Given that it will be
manually updated, we can assume it won't be too large to affect performance.
The mechanism to add and delete from the black list is done via a dynamically
generated module so that we don't have to constantly look up the ETS table.
Most of the code is ported over from couch_log. 

When and index is added to the black list while it is still indexing, nothing will 
prevent the spawned process from completing the indexing task.
This is expected as we currently do not want to add complexity
to the indexing code. If indexing process dies, it will respawn and
wait for a new trigger to spawn the updater process. However, if the black
list is in place, any new triggers will be ignored. In essence, we let
existing indexers to finish, but we don't allow new ones to start once the index
is added to the black list.

When a black listed indexed is part of a search, the user receives a 400 with
the database name, design doc id, and index name that notifies him the index is
blacklisted.

If the user tries to delete and recreate the index with the same name, it will
still be black listed. Once the user modifies the index definition, and notifies
our operator, the operator can delete the index from the list. The index will
begin updating on the next update to the db or from a new search request.

BugzId: 109229


src/dreyfus_util.erl
```diff
@@ -166,6 +168,71 @@ time(Metric, {M, F, A}) when is_list(Metric) ->
...
+    case in_black_list(DbName, GroupId, IndexName) of
+        true ->
+            Reason = ?l2b(io_lib:format("Index <~s, ~s, ~s>, is BlackListed",
+                [?b2l(DbName), ?b2l(GroupId), ?b2l(IndexName)])),
+            throw ({bad_request, Reason});
+        _ ->
+            ok
+    end.
+
+add_bl_element(DbName, GroupId, IndexName) when is_list(DbName),
```
jiangphcn: I didn't see that `add_bl_element/3` and `add_bl_element/1` are called in production codes (they are called in test code for sure). I can understand that they can be used for operator to add entry to blacklist. Just wonder whether there is need to have meaning response to operators and let them know whether entry is added or not?

src/dreyfus_index.erl
```diff
@@ -120,23 +120,34 @@ init({DbName, Index}) ->
...
+        false ->
+            UpPid = spawn_link(fun() ->
+                dreyfus_index_updater:update(IndexPid,Index)
+            end),
+            State#state{
+                updater_pid=UpPid,
+                waiting_list=[{From,RequestSeq}|WaitList]
+            };
+        _ ->
+            couch_log:notice("Index Blocked from Updating - db: ~p, ddocid: ~p name: ~p", [DbName, DDocId, IdxName]),
```
jiangphcn: more than 80 characters

src/dreyfus_index.erl
```diff
@@ -120,23 +120,34 @@ init({DbName, Index}) ->
...
+            State
+    end,
+    {noreply, NewState};
 handle_call({await, RequestSeq}, _From,
             #state{index=#index{current_seq=Seq}}=State) when RequestSeq =< Seq ->
     {reply, {ok, State#state.index_pid, Seq}, State};
 handle_call({await, RequestSeq}, From, #state{waiting_list=WaitList}=State) ->
-    {noreply, State#state{
```
jiangphcn, line 137: why `noreply` is removed?

jiangphcn (commented):
Looks good and leave several comments. Will check test codes later.

src/dreyfus_config.erl
```diff
@@ -0,0 +1,69 @@
+-module(dreyfus_config).
```
iilyak: Did you consider using "data providers" feature from `couch_epi`?
It generates module for you and it regenerates it when it is updated.
It would also use one function clause per blacklisted element. Effectively giving you hashing structure your mentioned in the description of the PR.

Here is a quick and dirty manual test to illustrate how it works:
```
1> Handle = couch_epi_data_gen:get_handle({dreyfus, config}).
couch_epi_data_gen_dreyfus_black_list
2> couch_epi_data_gen:generate(Handle, [
    {black_list, [
        {["mydb", "myddocid", "myindexname"], true},
        {["mydb2", "myddocid2", "myindexname2"], true}
    ]}
]).
ok
3>couch_epi_data_gen:get(Handle, black_list, ["mydb", "myddocid", "myindexname"]).
true
4>couch_epi_data_gen:get(Handle, black_list, ["mydb2", "myddocid2", "myindexname2"]).
true
5> couch_epi_data_gen:get(Handle, black_list, ["mydb_miss", "myddocid", "myindexname"]).
undefined
```

In order to plug this feature properly to your code you would need to.
1. define data provider module
   ```
    -module(dreyfus_config).
    -export([data/0]).
    data() ->
         [{K, true} || K <- config:get("dreyfus", "black_list", [])].
   ```

2. Add data provider to epi config module (the one which implements `couch_epi_plugin` behaviour)
    ```
    data_providers() ->
       [
           {{dreyfus, black_list}, {module, dreyfus_config}, [{interval, 100}]}
       ].
    ```

3. If you want to get notification when blacklist changes you would need to also define a subscriber for the events and notify callback function
    ```
    data_subscriptions() ->
         [{dreyfus, black_list}].

    notify(Key, OldValue, NewValue) ->
         % do something here
         ok. 
    ```

This would eliminate most of the code in the PR.

src/dreyfus_util.erl
tonysun83: @jiangphcn : we need to to add or remove from the list. otherwise, the operator has to call ```config:set/3``` for an entire new list. Remember, config doesn't have adding or removing from list capabilities.

src/dreyfus_index.erl
tonysun83: it's at the bottom

src/dreyfus_config.erl
tonysun83: @iilyak : From my understanding, ```data/0``` inside the ```dreyfus_config``` module is constantly being called. I set the interval to 2.5 seconds, and my debug output shows ```config:get``` being called inside ```data/0``` every 2.5 seconds based on the code above. If ```data/0``` returns a different value, then the module should be regenerated right? Unfortunately, I am not seeing this behavior happen. 

In the supervisor init I added a ```dreyfus_config:init/0 ``` call

```
init(_Args) ->
    ok = dreyfus_config:init(),
    Children = [
        child(dreyfus_index_manager)
    ],
    {ok, {{one_for_one,10,1},
        couch_epi:register_service(dreyfus_epi, Children)}}.
```
This simply does:
```
init() ->
    Handle = couch_epi_data_gen:get_handle({dreyfus, config}),
    couch_epi_data_gen:generate(Handle, [{black_list, []}]).
```
This I would assume generates the module```couch_epi_data_gen_dreyfus_black_list``` with no function clauses since the list is empty.
When I call ```config:set/3```,  I see the new value appear in ```data/0```, but when
I call ```couch_epi_data_gen:get(Handle, black_list, Key).```, the value is still undefined.
I think the module is not being regenerated to contain a clause for the new Key. I even tried
on in a remsh session

```
(node1@127.0.0.1)1>  Handle = couch_epi_data_gen:get_handle({dreyfus, config}).
couch_epi_data_gen_dreyfus_config
(node1@127.0.0.1)2> couch_epi_data_gen:generate(Handle, [
(node1@127.0.0.1)2>     {black_list, [
(node1@127.0.0.1)2>         {["mydb", "myddocid", "myindexname"], true},
(node1@127.0.0.1)2>         {["mydb2", "myddocid2", "myindexname2"], true}
(node1@127.0.0.1)2>     ]}
(node1@127.0.0.1)2> ]).
ok
(node1@127.0.0.1)3> couch_epi_data_gen:get(Handle, black_list, ["mydb", "myddocid", "myindexname"]).
true
(node1@127.0.0.1)4> rpc:multicall(config, set, ["dreyfus", "black_list", [["testdb1", "6a9805a78876be4477e00de79b972a413059efae", "index-all"]]]). 
{[ok,ok,ok],[]}
(node1@127.0.0.1)5> 
(node1@127.0.0.1)5> couch_epi_data_gen:get(Handle, black_list, ["testdb1", "6a9805a78876be4477e00de79b972a413059efae", "index-all"]).
undefined
(node1@127.0.0.1)6> couch_epi_data_gen:get(Handle, black_list, ["mydb2", "myddocid2", "myindexname2"]).
true
```

I wonder if my approach is somewhat flawed.

src/dreyfus_index.erl
iilyak: It looks like all 3 arguments are strings maybe we should use `'~s'`? Not a big deal.

src/dreyfus_config.erl
tonysun83: @iilyak : I cherry-picked your commits from https://github.com/apache/couchdb/pull/1432/files but I am still getting a similar behavior as above. The difference now is that notify is being called.

tonysun83:
This PR depends on https://github.com/apache/couchdb/pull/1432

tonysun83:
@iilyak @jiangphcn I updated to use couch_epi. 1 second polling shouldn't affect performance too much, but will need to do performance tests

src/dreyfus_index.erl
jiangphcn: The diff for this clause looks: 
```
 handle_call({await, RequestSeq}, From, #state{waiting_list=WaitList}=State) ->
-    {noreply, State#state{
+    State#state{
         waiting_list=[{From,RequestSeq}|WaitList]
-    }};
+    };
```
Please remind me where is `noreply`.

src/dreyfus_util.erl
jiangphcn: So `add_bl_element/3` will be called by operator to add or remove entity from list, right? My comment is whether we need to have meaningful return code/message to operator so that they know the actual result.

src/dreyfus_index.erl
tonysun83: @jiangphcn look at the entire file, line 144.

src/dreyfus_util.erl
tonysun83: @jiangphcn : I understand your idea. Perhaps we can move this over to snippet with a more meaningful response instead since it is not used in the source code?

src/dreyfus_index.erl
tonysun83: @jiangphcn : nevermind, not sure why that was changed, okay I see it now

src/dreyfus_index.erl
```diff
@@ -146,9 +146,9 @@ handle_call({await, RequestSeq}, _From,
             #state{index=#index{current_seq=Seq}}=State) when RequestSeq =< Seq ->
     {reply, {ok, State#state.index_pid, Seq}, State};
 handle_call({await, RequestSeq}, From, #state{waiting_list=WaitList}=State) ->
-    State#state{
+    {no_reply, State#state{
```
jiangphcn: :+1:

src/dreyfus_config.erl
```diff
@@ -0,0 +1,11 @@
+ -module(dreyfus_config).
+
+ -export([data/0, get/1]).
+
+data() ->
+    Data = config:get("dreyfus", "black_list", []),
```
iilyak: I thought about it over the weekend and think that we shouldn't do that.
This configuration gives as a number of complications:
1. the operators  have to use API and cannot use regular config
   - which means that tooling wouldn't be able to use `_config` endpoint
   - complex term would be stored in the config
   - there is a possibility that operator would use config by mistake and overwrite the rules.
2. it requires complex `in_black_list/3` and `add_bl_element/1` functions

It would simplify things a lot if we change the way we configure the blacklist somehow.

tonysun83: @iilyak :  @jiangphcn  had the same concern with the complex functions. What do you think about moving the complex add/remove over into snippet? We would still keep ```in_black_list/3``` but it would not be called by any operators. They can obviously still update the config manually by doing ```rpc:call/3``` or ```config:set/3```.

jiangphcn: @tonysun83 I think that it would be better to move the complex add/remove to snippet. Operators are more familiar with snippet than dbcore codes. Thus, they may easily use the snippet to add/remove entry from blacklist.

test/dreyfus_test_util.erl
```diff
@@ -55,3 +56,48 @@ get_handler() ->
...
+            BlackList;
+        false ->
+            NewList = [IndexEntry | BlackList],
+            config:set("dreyfus", "black_list", NewList),
+            NewList
+    end,
+    UpdatedList;
+add_bl_element(_IndexEntry) ->
+    config:get("dreyfus", "black_list", []).
+
```
tonysun83: whitespace

src/dreyfus_config.erl
tonysun83: @jiangphcn : i've moved the functions over to test utils. They will also be added to snippet

iilyak: @tonysun83: What if we would replace a single configuration key with a section. every rule would get own key. For example:
```
[dreyfux_blacklist]
mydb3/_design/myddocid3/myindexname3 = true
```
The code change to this PR would be 2-3 lines. However it wouldn't be possible for operator to make a mistake. Also there is a possibility that the direct remsh access would go away (security and compliance requirements) so snippets wouldn't be a solution.

tonysun83:
@iilyak @jiangphcn : I modified the code to allow a new section for individual indexes. This should allow us to rely on config only and not external code.

test/dreyfus_blacklist_await_test.erl
```diff
@@ -0,0 +1,77 @@
...
+        Msg = "Index Blocked from Updating - db: ~p, ddocid: ~p name: ~p",
+        Return = wait_log_message(Msg, fun() ->
+            {noreply, NewState} = dreyfus_index:handle_call({await, 1},
+            self(), State)
+        end),
+        ?assertEqual(Return, ok)
+    end).
+
+
+wait_log_message(Fmt, Fun) ->
```
iilyak: This is awesome!!!

test/dreyfus_test_util.erl
```diff
@@ -4,100 +4,10 @@
...
-            NewList = lists:delete(IndexEntry, BlackList),
-            config:set("dreyfus", "black_list", NewList),
-            NewList;
-        false ->
-            BlackList
-    end,
-    UpdatedList;
-remove_bl_element(_IndexEntry) ->
-    config:get("dreyfus", "black_list", []).
+wait_config_change(Key, Value) ->
```
iilyak: Nice!!!

test/dreyfus_test_util.erl
```diff
@@ -4,100 +4,10 @@
 
 -include_lib("couch/include/couch_db.hrl").
 
-add_listener(Listener) ->
```
iilyak: I love when code is removed. Good job!

jiangphcn:
LGTM, Nice work, Tony.

test/dreyfus_blacklist_request_test.erl
```diff
@@ -56,18 +56,16 @@ deny_fabric_requests() ->
...
-        ?assertThrow({bad_request, Reason}, dreyfus_fabric_group1:go(<<"mydb">>,
-            <<"myddocid">>,  <<"myindexname">>, QueryArgs)),
-        ?assertThrow({bad_request, Reason}, dreyfus_fabric_group2:go(<<"mydb">>,
-            <<"myddocid">>,  <<"myindexname">>, QueryArgs)),
-        ?assertThrow({bad_request, Reason}, dreyfus_fabric_info:go(<<"mydb">>,
-            <<"myddocid">>,  <<"myindexname">>, QueryArgs))
-    end).
+    config:set("dreyfus_blacklist", Index, "true"),
+    dreyfus_test_util:wait_config_change(Index, "true"),
+    ?assertThrow({bad_request, Reason}, dreyfus_fabric_search:go(<<"mydb">>,
```
iilyak: I am having the following failure:
```
    dreyfus_blacklist_request_test: dreyfus_blacklist_request_test_...*failed*
in function dreyfus_blacklist_request_test:'-deny_fabric_requests/0-fun-0-'/0 (test/dreyfus_blacklist_request_test.erl, line 61)
in call from dreyfus_blacklist_request_test:deny_fabric_requests/0 (test/dreyfus_blacklist_request_test.erl, line 61)
**error:{assertException,
    [{module,dreyfus_blacklist_request_test},
     {line,61},
     {expression,
         "dreyfus_fabric_search : go ( << \"mydb\" >> , << \"myddocid\" >> , << \"myindexname\" >> , QueryArgs )"},
     {pattern,"{ throw , { bad_request , Reason } , [...] }"},
     {unexpected_exception,
         {error,database_does_not_exist,
             [{mem3_shards,load_shards_from_db,"mydb",[{file,...},{...}]},
              {mem3_shards,load_shards_from_disk,1,[{...}|...]},
              {mem3_shards,for_db,2,[...]},
              {dreyfus_fabric_search,go,4,...},
              {dreyfus_blacklist_request_test,
                  '-deny_fabric_requests/0-fun-0-',...},
              {dreyfus_blacklist_request_test,...},
              {...}|...]}}]}
  output:<<"">>
```

It seems there are two problems here:
1. the `dreyfus_util:maybe_deny_index/4` is only called when `dreyfus_fabric_search:go/4` matches `is_binary(GroupId)`
2. The mock is not sufficient.

test/dreyfus_blacklist_request_test.erl
```diff
@@ -76,10 +76,10 @@ allow_fabric_request() ->
     ok = meck:expect(dreyfus_fabric_search, go, fun(A, DDocId, B, C) ->
```
iilyak: Please fix compiler warnings here:
```
/home/jenkins/src/dreyfus/test/dreyfus_blacklist_request_test.erl:76: Warning: variable 'A' is unused
/home/jenkins/src/dreyfus/test/dreyfus_blacklist_request_test.erl:76: Warning: variable 'B' is unused
/home/jenkins/src/dreyfus/test/dreyfus_blacklist_request_test.erl:76: Warning: variable 'C' is unused
/home/jenkins/src/dreyfus/test/dreyfus_blacklist_request_test.erl:76: Warning: variable 'DDocId' is unused
```

iilyak (approved):
+1 after squash. Great work !!!
```
======================== EUnit ========================
module 'dreyfus_test_util'
module 'dreyfus_index'
module 'dreyfus_epi'
module 'dreyfus_fabric_info'
  [done in 0.001 s]
module 'dreyfus_fabric_cleanup'
dreyfus_fabric_search: merge_facets_test (module 'dreyfus_fabric_search')...ok
module 'clouseau_rpc'
module 'dreyfus_fabric_group1'
module 'dreyfus_blacklist_request_test'
  dreyfus blacklist request tests
Application crypto was left running!
    dreyfus_blacklist_request_test: dreyfus_blacklist_request_test_...[0.582 s] ok
    dreyfus_blacklist_request_test: dreyfus_blacklist_request_test_...[0.991 s] ok
    [done in 1.580 s]
  [done in 7.115 s]
module 'dreyfus_httpd'
module 'dreyfus_httpd_handlers'
module 'dreyfus_config_test'
  dreyfus config tests
Application asn1 was left running!
Application b64url was left running!
Application compiler was left running!
Application crypto was left running!
Application khash was left running!
Application folsom was left running!
Application couch_stats was left running!
Application hyper was left running!
Application hqueue was left running!
Application ioq was left running!
Application public_key was left running!
Application syntax_tools was left running!
Application xmerl was left running!
Application mochiweb was left running!
Application sasl was left running!
Application os_mon was left running!
[os_mon] memory supervisor port (memsup): Erlang has closed
[os_mon] cpu supervisor port (cpu_sup): Erlang has closed
Application rexi was left running!
Application mem3 was left running!
Application dreyfus was left running!
    dreyfus_config_test: dreyfus_config_test_...[0.945 s] ok
    dreyfus_config_test: dreyfus_config_test_...[0.959 s] ok
[os_mon] memory supervisor port (memsup): Erlang has closed
[os_mon] cpu supervisor port (cpu_sup): Erlang has closed
    [done in 1.910 s]
  [done in 2.132 s]
module 'dreyfus_fabric_group2'
module 'dreyfus_sup'
module 'dreyfus_app'
module 'dreyfus_rpc'
module 'dreyfus_blacklist_await_test'
  dreyfus black_list_doc await tests
    dreyfus_blacklist_await_test: dreyfus_blacklist_await_test_...test/dreyfus_blacklist_await_test.erl:33:<0.872.0>: config: [dreyfus_blacklist] mydb.black_list_doc.my_index set to true for reason nil
test/dreyfus_blacklist_await_test.erl:33:<0.1165.0>: Index Blocked from Updating - db: <<"mydb">>, ddocid: <<"_design/black_list_doc">> name: <<"my_index">>
[0.128 s] ok
    [done in 0.193 s]
  [done in 0.405 s]
module 'dreyfus_util'
  dreyfus_util: empty_test...ok
  dreyfus_util: primary_asc_test...ok
  dreyfus_util: primary_desc_test...ok
  dreyfus_util: secondary_asc_test...ok
  dreyfus_util: secondary_desc_test...ok
  dreyfus_util: stash_test...[0.001 s] ok
  [done in 0.019 s]
module 'dreyfus_config'
module 'dreyfus_index_updater'
module 'dreyfus_bookmark'
module 'dreyfus_fabric'
module 'dreyfus_index_manager'
=======================================================
  All 12 tests passed.
```

PR #28, opened 2018-07-26
Partition key search

garrensmith:
This is the search implementation for the new partition work we doing for CouchDB.

### DIFF
--- a/src/dreyfus_httpd.erl
+++ b/src/dreyfus_httpd.erl
@@ -16,11 +16,29 @@
 -module(dreyfus_httpd).
 
 -export([handle_search_req/3, handle_info_req/3, handle_disk_size_req/3,
-         handle_cleanup_req/2, handle_analyze_req/1]).
+         handle_cleanup_req/2, handle_analyze_req/1, handle_partition_search_req/4]).
 -include("dreyfus.hrl").
 -include_lib("couch/include/couch_db.hrl").
 -import(chttpd, [send_method_not_allowed/2, send_json/2, send_json/3,
                  send_error/2]).
+
+
+handle_partition_search_req(#httpd{method=Method,
+                            path_parts=[_, _, _, _, _, _, IndexName]} = Req,
+                            Db, DDoc, Partition)
+  when Method == 'GET'; Method == 'POST' ->
+    QueryArgs = parse_index_params(Req),
+    QueryArgs1 = QueryArgs#index_query_args{
+        partition = Partition,
+        partitioned = true
+    },
+    check_partition_restrictions(QueryArgs1),
+    handle_search_req_int(Req, Db, DDoc, IndexName, QueryArgs1, 0, 500);
+handle_partition_search_req(#httpd{path_parts=[_, _, _, _, _]}=Req, _Db, _DDoc, _Partition) ->
+    send_method_not_allowed(Req, "GET,POST");
+handle_partition_search_req(Req, _Db, _DDoc, _Partition) ->
+    send_error(Req, {bad_request, "path not recognized"}).
+
 
 handle_search_req(Req, Db, DDoc) ->
     handle_search_req(Req, Db, DDoc, 0, 500).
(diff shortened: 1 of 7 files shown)


src/dreyfus_httpd.erl
```diff
@@ -495,3 +510,64 @@ backoff_and_retry(Req, Db, DDoc, RetryCount, RetryPause) ->
...
+        grouping = Grouping,
+        drilldown = Drilldown,
+        ranges = Ranges
+    } = Args,
+
+    #grouping{
+        by = GroupingBy
+    } = Grouping,
+
+    case PartitionKey of
```
garrensmith: I feel like there is a better way of doing all these checks. I also considered doing something like adding each field we checking into a list that contains the value it should equal and error message. Then looping over and checking the values, instead of all these case statements.

src/dreyfus_httpd.erl
```diff
@@ -35,6 +35,15 @@ handle_search_req(#httpd{method=Method, path_parts=[_, _, _, _, IndexName]}=Req
         include_docs = IncludeDocs,
         grouping = Grouping
     } = parse_index_params(Req),
+    QueryArgs1 = QueryArgs#index_query_args{
+        partitioned = mem3:is_partitioned(hd(mem3:shards(DbName))) % hideous
```
tonysun83: if there's a dependency PR for ```mem3:is_partitioned```, perhaps we reference it in this PR?

src/dreyfus_httpd.erl
```diff
@@ -495,3 +510,64 @@ backoff_and_retry(Req, Db, DDoc, RetryCount, RetryPause) ->
...
+            ok;
+        _ ->
+            partition_error(<<"Cannot use count in a partition query">>)
+    end,
+
+    case Drilldown of 
+        [] ->
+            ok;
+        _ ->
+            partition_error(<<"Cannot use count in a partition query">>)
```
tonysun83: drilldown

src/dreyfus_httpd.erl
```diff
@@ -495,3 +510,64 @@ backoff_and_retry(Req, Db, DDoc, RetryCount, RetryPause) ->
...
+    end,
+
+    case Ranges of 
+        nil ->
+            ok;
+        _ ->
+            partition_error(<<"Cannot use count in a partition query">>)
```
tonysun83: ranges

src/dreyfus_httpd.erl
tonysun83: you can probably create the list and run a ```lists:foreach``` on it to get the desired looping effect

src/dreyfus_httpd.erl
```diff
@@ -245,6 +254,8 @@ validate_index_query(highlight_number, Value, Args) ->
     Args#index_query_args{highlight_number=Value};
 validate_index_query(highlight_size, Value, Args) ->
     Args#index_query_args{highlight_size=Value};
+validate_index_query(partition, Value, Args) ->
```
tonysun83: partitioned?

src/dreyfus_httpd.erl
```diff
@@ -294,6 +305,8 @@ parse_index_param("highlight_number", Value) ->
     [{highlight_number, parse_positive_int_param2("highlight_number", Value)}];
 parse_index_param("highlight_size", Value) ->
     [{highlight_size, parse_positive_int_param2("highlight_size", Value)}];
+parse_index_param("partition", Value) ->
```
tonysun83: partitioned?

src/dreyfus_httpd.erl
```diff
@@ -339,6 +352,8 @@ parse_json_index_param(<<"highlight_number">>, Value) ->
     [{highlight_number, parse_positive_int_param2("highlight_number", Value)}];
 parse_json_index_param(<<"highlight_size">>, Value) ->
     [{highlight_size, parse_positive_int_param2("highlight_size", Value)}];
+parse_json_index_param(<<"partition">>, Value) ->
```
tonysun83: partitioned?

src/dreyfus_index_updater.erl
```diff
@@ -64,7 +64,7 @@ update(IndexPid, Index) ->
         couch_db:close(Db)
     end.
 
-load_docs(FDI, {I, IndexPid, Db, Proc, Total, LastCommitTime}=Acc) ->
+load_docs(FDI, {I, IndexPid, Db, Proc, Total, LastCommitTime, DbName}=Acc) ->
```
tonysun83: do we need to pass in DbName here? would `couch_db:name(Db)` return what's needed?

src/dreyfus_httpd.erl
```diff
@@ -495,3 +510,64 @@ backoff_and_retry(Req, Db, DDoc, RetryCount, RetryPause) ->
...
     end.
+
+
+partition_error(Msg) ->
+    throw({search_partition_error, Msg}).
+
+
+check_partition_restrictions(#index_query_args{} = Args) ->
+    #index_query_args{
+        partition_key= PartitionKey,
```
tonysun83: need a space before =, same goes for the line below

src/dreyfus_index_updater.erl
```diff
@@ -78,14 +78,27 @@ load_docs(FDI, {I, IndexPid, Db, Proc, Total, LastCommitTime}=Acc) ->
...
         false ->
             {ok, setelement(1, Acc, I+1)}
     end.
+
+
+maybe_add_partition_key(Db, Id, Fields) ->
+    case mem3:is_partitioned(Db) of 
+        true ->
+            PK = hd(binary:split(Id, <<":">>)),
+            Fields1 = [{<<"_partition">>, PK, {[]}} | Fields],
```
tonysun83: If I remember correctly, the tuples were just {Key, Value} pairs that were going to be indexed. So for partitioning work, you are adding a new field ```<<"_partition">>``` with value ```PK```. What is the ```{[]}``` for (Might be options of I'm forgetting)?

garrensmith:
Currently the main Partition work is happening over here https://github.com/apache/couchdb/tree/user-partitioned-dbs-wip

src/dreyfus_httpd.erl
garrensmith: Good point. I've mented in the comments but for quick reference https://github.com/apache/couchdb/tree/user-partitioned-dbs-wip

src/dreyfus_index_updater.erl
garrensmith: I tried that and it didn't work. But the is db partitioned work in mem3 is likely to change. So I'll update the branch once it does.

garrensmith:
@tonysun83 thanks for the review. I've made a ton of changes based on your feedback.

garrensmith:
closing due to #34

PR #32, opened 2018-10-25
use updated_on instead of timestamp_utc in local purge doc

jiangphcn:
Use `updated_on` instead of `timestamp_utc` in local purge doc

COUCHDB-3326


davisp (approved):
+1

PR #34, opened 2018-12-04
Implement partitioned queries

garrensmith:
This is the dreyfus work for the new partition queries.

PR #36, opened 2018-12-04
Database partitions tests

garrensmith:
The main part is adding in the partition search tests. I had to add some validation in so that the tests passed. So I added validation so that a non-partitioned design doc cannot be used in a
partition query and that a partitioned design doc cannot be used in a
global query


davisp (changes requested):
I think that the elixir suite should probably live under the test directory so that it mirrors CouchDB.

I'm also not seeing how this would be invoked via our root level Makefile without modifying it to reference `src/dreyfus/test/elixir/run` which seems not good. I wonder if there's a way to hook into rebar's eunit or some other thing to run these maybe? That's more of a nice to have. If there's nothing immediately obvious then hardcoding the path to `run` isn't the end of the world I don't think.

davisp:
Also, I'm not 100% sure that its good to restrict global queries on partitioned ddocs. Unless @rnewson can think of a reason it'd return invalid results I don't see a reason to force someone to define two indexes if the same one would work both places.

For views it makes sense because the returned results are incorrect. But here its just dropping a restriction on the _partition field in Lucene so I'd not reject it I don't think. Granted I'd also not expect people to rely on this I guess.

davisp:
This work is now included in #34.

PR #39, opened 2018-12-14
Fix function_clause caused by malformed accumulator

rnewson:
In a previous commit we added another item to the accumulator used by
load_docs. Unfortunately one clause was not modified so we send the
original 6-tuple, which then fails to match the function and crashes.

This affects searches of busy indexes, the user gets a function_clause
erorr instead of search results. It does not appear to prevent index
commits, though.

BugzID: 114420


nickva (approved):
+1

In general, once acc gets to 7 elements it's probably time for an #acc record...

rnewson:
agreed on record but I didn't want to get into that for an urgent patch.

PR #40, opened 2019-01-28
Fix search with limit using POST method

jiangphcn:
## Overview

Fix search with limit using POST method.

## Testing recommendations

mix test --trace test/partition_search_test.exs
```
Excluding tags: [pending: true]


PartitionSearchTest
  * test normal search on non-partitioned dbs with over limit (824.5ms)
  * test normal search on non-partitioned dbs without limit (1386.5ms)
  * test Cannot do global query with partition view (1021.2ms)
  * test Cannot do partition query with global search ddoc (774.2ms)
  * test Only returns docs in partition not those in shard (1336.5ms)
  * test Works with limit using POST for partitioned db (1036.9ms)
  * test Works with limit using POST for on non-partitioned db (1296.3ms)
  * test normal search on non-partitioned dbs with limit (1360.3ms)
  * test Simple query returns partitioned search results (1587.4ms)
  * test normal search on non-partitioned dbs still work (1331.2ms)
  * test Works with bookmarks and limit (1437.1ms)

PartitionHelpers


Finished in 14.9 seconds
11 tests, 0 failures

Randomized with seed 390897
```

## Checklist

- [X] Code is written and works correctly;
- [X] Changes are covered by tests;
- [ ] Documentation reflects the changes;

### DIFF
--- a/src/dreyfus_httpd.erl
+++ b/src/dreyfus_httpd.erl
@@ -312,7 +312,7 @@ parse_json_index_param(<<"bookmark">>, Value) ->
 parse_json_index_param(<<"sort">>, Value) ->
     [{sort, Value}];
 parse_json_index_param(<<"limit">>, Value) ->
-    [{limit, ?JSON_DECODE(Value)}];
+    [{limit, Value}];
 parse_json_index_param(<<"stale">>, <<"ok">>) ->
     [{stale, ok}];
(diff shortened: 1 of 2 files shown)


davisp (approved):
+1

tonysun83 (approved):

garrensmith (approved):

PR #41, opened 2019-02-05
Partition improvements

garrensmith:
Improves the error messages to better explain why an error has occurred.
Fixes issue where a partition query using post would not work.
Checks if the partition field is added in the body and url
for a POST search request

### DIFF
--- a/src/dreyfus_httpd.erl
+++ b/src/dreyfus_httpd.erl
@@ -189,8 +189,15 @@ parse_index_params(#httpd{method='GET'}=Req, Db) ->
         chttpd:qs(Req)),
     parse_index_params(IndexParams, Db);
 parse_index_params(#httpd{method='POST'}=Req, Db) ->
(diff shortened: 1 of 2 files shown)


davisp (approved):
+1

src/dreyfus_httpd.erl
```diff
@@ -424,6 +431,19 @@ parse_non_negative_int_param(Name, Val, Prop, Default) ->
     end.
 
 
+ensure_unique_partition(IndexParams) ->
+    Partitions = lists:filter(fun({Key, _Val}) ->
+        Key == partition
+    end, IndexParams),
+    case length(length:usort(Partitions)) > 1 of
```
rnewson, line 438: you mean lists:usort, surely?

davisp: Surely. Fixed.

PR #42, opened 2019-02-13
fix function_clause for partition queries

tonysun83:
JsonBody was already a tuple list so putting in brackets made it
a nested list which function claused on {K,V}. This fixes that.


davisp (approved):
+1

PR #46, opened 2019-04-12
Use dedicated search IOQ channel for

chewbranca:
This switches to using a dedicated `search` IOQ channel for all search operations, rather than piggy-backing off of the `interactive` and `view_update` channels. Switching these operations to a dedicated search channel allows us to bypass search from IOQ in isolation, rather than having to bypass both interactive and view update.

We could be more granular and declare a `search` channel and a `search_update` channel, but I'm not sure if that actually matters here.

This PR also adds a metric declaration for search traffic in the `couchdb.io_queue2` metrics namespace. It's somewhat awkward to add that metric here, but if we're not going to introduce the dynamic channels in couchdb-ioq, then we need to declare the metrics somewhere.

### DIFF
--- a/priv/stats_descriptions.cfg
+++ b/priv/stats_descriptions.cfg
@@ -51,3 +51,9 @@
     {type, histogram},
     {desc, <<"length of an dreyfus_index info request">>}
 ]}.
+
+%% Declare IOQ2 search channel metrics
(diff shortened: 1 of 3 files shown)


davisp (approved):
+1

PR #47, opened 2019-05-10
adjust metric for io_queue-search

jiangphcn:
This PR is to eliminate `unknown metric: [couchdb,io_queue,search]` emission.

### DIFF
--- a/priv/stats_descriptions.cfg
+++ b/priv/stats_descriptions.cfg
@@ -52,6 +52,12 @@
     {desc, <<"length of an dreyfus_index info request">>}
 ]}.
 
+%% Declare IOQ search channel metrics
+{[couchdb, io_queue, search], [
+    {type, counter},
+    {desc, <<"Search IO directly triggered by client requests">>}
+]}.
+
 %% Declare IOQ2 search channel metrics
 {[couchdb, io_queue2, search, count], [
     {type, counter},


priv/stats_descriptions.cfg
```diff
@@ -52,8 +52,14 @@
     {desc, <<"length of an dreyfus_index info request">>}
 ]}.
 
+%% Declare IOQ search channel metrics
+{[couchdb, io_queue, search], [
+    {type, counter},
+    {desc, <<"Search IO directly triggered by client requests">>}
+]}.
+
```
chewbranca, line 60: Good catch, I forgot to add the associated metric for IOQ1. This metric is updated by https://github.com/apache/couchdb-ioq/blob/master/src/ioq_server.erl#L424 and we can see it's of the format `[couchdb, io_queue, search]` and this change is correct.

priv/stats_descriptions.cfg
```diff
@@ -52,8 +52,14 @@
...
 %% Declare IOQ2 search channel metrics
-{[couchdb, io_queue2, search, count], [
+{[couchdb, io_queue2, search], [
```
chewbranca: This is the corresponding search metric for IOQ2 and it should not have the format changed. You can see in https://github.com/apache/couchdb-ioq/blob/master/src/ioq_server2.erl#L451 that for IOQ2 class counters we're using the form `[couchdb, io_queue2, Class, count]`, so this change should be reverted unless you've specifically seen an issue with this. Your comment suggests that the unknown metric errors are _only_ for the IOQ1 case: `unknown metric: [couchdb,io_queue,search]`.

chewbranca (changes requested):

jiangphcn:
Thank you @chewbranca and @nickva. Good catch, the change on `[couchdb, io_queue2, search]` was reverted. Would you please approve this PR if there is no other comment?

nickva (approved):
@chewbranca thanks for taking a look. I had missed the part about ioq2 as well.

@jiangphcn nice, fix!

